PR #4, opened 2026-04-15
feat(server): add OpenAI-compatible /v1/audio/transcriptions endpoint…

ubaldus:
… with multi-format support

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
           submodules: recursive
 
       - name: Install deps
-        run: sudo apt-get update && sudo apt-get install -y cmake g++ libopenblas-dev
+        run: sudo apt-get update && sudo apt-get install -y cmake g++ libopenblas-dev patchelf
 
       - name: Build
         run: |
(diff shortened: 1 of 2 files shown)


CrispStrobe:
Thanks for the PR and for opening the door on OpenAI API compatibility — this is a feature we wanted!

After reviewing, we decided to implement it from scratch rather than iterate on this diff. The main reasons:

- **Reuse existing formatters** — `crispasr_output.{h,cpp}` already has SRT/VTT/JSON writers with correct centisecond-to-timestamp conversion. The PR duplicated these inline, which would drift over time.
- **Hardcoded `"language": "english"`** in `segments_to_openai_verbose` — needs to pass through the actual requested/detected language.
- **Temp file security** — `mkstemp()` instead of predictable timestamp-based filenames.
- **`std::stof` crash** — uncaught exception on malformed `temperature` field would crash the server.
- **Unit safety** — `crispasr_segment::t0/t1` are centiseconds; the PR assumed different units in different places.
- **Shared transcription pipeline** — extracted `do_transcribe()` so `/inference` and `/v1/audio/transcriptions` share the same audio-load + transcribe path instead of duplicating it.

The RPATH/patchelf release CI change is a good idea but belongs in a separate PR — it's orthogonal to the API endpoint.

We've pushed our implementation with attribution in the commit message. The endpoint is live at `POST /v1/audio/transcriptions` with all five response formats (`json`, `verbose_json`, `text`, `srt`, `vtt`), plus `GET /v1/models`.

Thanks again for the contribution idea!

CrispStrobe:
Closing in favor of our own implementation (see comment above). Thanks for the idea!

Issue #56, opened 2026-05-04
[Bug] qwen3-tts-1.7b-base: GGML_ASSERT crash with default O15 graph reuse on CUDA


vkrmch:
## Summary

After upgrading past commit `5e21e4a` (`perf(qwen3-tts): default O15 cp_pred graph reuse to ON`, 2026-05-02), Qwen3-TTS 1.7B-Base crashes on the first generation step with a GGML_ASSERT in the CUDA backend. Setting `QWEN3_TTS_O15=0` works around it, so the cached-graph code path appears CUDA-incompatible (or has a tensor-binding race) when used by default.

## Environment

- **Device:** NVIDIA Jetson Orin AGX 64GB
- **L4T / CUDA:** 12.8 (dustynv/pytorch base image's CUDA toolchain on host)
- **Compute capability:** sm_87 (built with `-DCMAKE_CUDA_ARCHITECTURES=87-real`)
- **Build:** `cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_CUDA=ON -DGGML_NATIVE=ON -DCMAKE_CUDA_ARCHITECTURES=87-real`
- **CrispASR HEAD:** `d74edf9`
- **Backend:** `qwen3-tts-1.7b-base`, model auto-resolved via `-m auto`

## Symptom

On the first call into `code_pred_generate_15`, the process aborts with:

```
GGML_ASSERT(tensor) failed
  in ggml_backend_tensor_set (ggml-backend.c)
```

This happens before any audio is emitted. With `QWEN3_TTS_O15=0`, the same invocation runs to completion (5.04s of audio at RTF ~0.86 on Orin AGX, voice clone working as expected).

## Reproducer

```bash
QWEN3_TTS_BENCH=1 ./build/bin/crispasr \
  --backend qwen3-tts-1.7b-base -m auto \
  --voice path/to/reference.wav \
  --ref-text \"<transcript of reference>\" \
  --tts \"Hello, this is a voice clone test.\" \
  --tts-output /tmp/out.wav
```

vs. the working invocation (note the env var):

```bash
QWEN3_TTS_O15=0 QWEN3_TTS_BENCH=1 ./build/bin/crispasr ...
```

## Suspect commits

- `7298dd5` (2026-04-30) — `fix(qwen3-tts): zero-init code_pred KV cache + make O15 path opt-in`
[...]

PR #57, opened 2026-05-04
perf(qwen3-tts): QWEN3_TTS_CODEC_GPU env var for clean GPU codec path

vkrmch:
## Summary

Adds a new env var `QWEN3_TTS_CODEC_GPU=1` that routes the qwen3-tts codec decode through the main GPU scheduler **without** the per-op trace callback that `QWEN3_TTS_CODEC_FORCE_METAL=1` installs. On non-Mac GPU platforms (CUDA / Vulkan / etc.) the M1 Metal hang the default CPU pin is guarding against does not apply, and codec on CPU is dramatically slower than codec on GPU — but the only existing switch pulls in a debugging trace callback that serialises every kernel and tanks throughput, so users on those platforms have no clean fast path today.

## Why

Today the codec is pinned to a CPU-only `codec_sched` by default. The comment in `qwen3_tts_context` explains why: Metal hangs on the codec graph on M1, so falling back to CPU was the safe choice. `QWEN3_TTS_CODEC_FORCE_METAL=1` was added as an escape hatch to reproduce that hang on demand for instrumentation — and as a side effect, it also routes the codec to `c->sched`, which is the only way today to get codec compute on the GPU at all.

But `FORCE_METAL` also gates a per-op `ggml_backend_synchronize` callback in `codec_decode_codes` (the same condition: `QWEN3_TTS_CODEC_TRACE || QWEN3_TTS_CODEC_FORCE_METAL`). That callback synchronises after every node — necessary for reproducing the Metal hang, but it serialises kernel launches and roughly triples codec wall time on CUDA.

So the current state is:
- **Default (CPU codec):** safe everywhere, but ~50× slower than GPU on Jetson Orin AGX in my testing.
- **`FORCE_METAL=1` (GPU + trace):** still much slower than it should be because of the per-op sync.
- **No way to get codec-on-GPU without the trace overhead.**

## What this changes

A sibling env var `QWEN3_TTS_CODEC_GPU=1` that:

1. Loads the codec weights onto `c->backend` instead of `c->backend_cpu`.
2. Makes `codec_pick_sched` return `c->sched` (main GPU sched).
3. Does NOT enable the trace callback — that stays gated on `QWEN3_TTS_CODEC_TRACE` / `QWEN3_TTS_CODEC_FORCE_METAL`.

`FORCE_METAL` semantics are unchanged so the M1-debug workflow is preserved exactly. The new var is documented in `docs/tts.md` alongside the existing qwen3-tts env switches and explicitly contrasted with `FORCE_METAL`.

## Measured impact

On Jetson Orin AGX (compute capability 8.7, CUDA 12.8), Qwen3-TTS-1.7B-Base Q8_0 + the matching 12 Hz tokenizer, voice-clone synthesis through `qwen3_tts_synthesize_codes` + `qwen3_tts_decode_codes`:

| Mode | Codec wall time per request | Notes |
|---|---|---|
| Default (CPU codec) | ~60 s for 2 s of audio | RTF ≫ 1, dominated by CPU |
| `FORCE_METAL=1` (GPU + trace) | ~17 s for 2 s of audio | Trace callback serialises kernels |
| `CODEC_GPU=1` (GPU, no trace) | ~1.5 s for 2 s of audio | RTF ~0.75 for the codec stage |

End-to-end RTF on this hardware drops from ~9 → ~1.5 with this flag plus the public `synthesize_codes` + `decode_codes` two-call pattern.

## Testing

- Built with `-DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=87-real` on the Jetson and ran end-to-end voice clone synthesis through `qwen3-tts-1.7b-base`. Audio output is identical between `FORCE_METAL=1` and `CODEC_GPU=1` modulo the trace overhead.
- I have not exercised the Vulkan path. If the codec graph also runs cleanly on Vulkan, `CODEC_GPU=1` should help there too — happy to test if you can point me at a known-good config.

## Out of scope (future PRs)

Two related findings I hit while debugging this — happy to file as separate issues / PRs if useful:

1. The bundled `qwen3_tts_synthesize` concatenates `ref_codes ++ gen_codes` before codec decode and trims the ref portion off the output. With a 26 s reference WAV (334 codec frames) this added a roughly constant ~16 s of codec compute per request on Jetson regardless of how much new audio was generated. Calling `qwen3_tts_synthesize_codes` + `qwen3_tts_decode_codes` directly produces clean audio on this hardware and avoids the constant cost. There may be a clean way to expose a "skip ref" knob inside `synthesize` itself.
2. The May 2 commits flipped `QWEN3_TTS_O15` to default-on, which crashes on first call into `code_pred_generate_15` on CUDA — already filed as #56.

Thanks for the project — pure C++ + ggml on Jetson is a great fit and this change makes it practical.

### DIFF
--- a/docs/tts.md
+++ b/docs/tts.md
@@ -142,6 +142,7 @@ defaults reproduce the validated, end-to-end-tested code path.
 | `QWEN3_TTS_PROF` | unset | Per-op profiler (more granular than `BENCH`). |
 | `QWEN3_TTS_CP_BACKEND` | unset | Pin the code predictor to a chosen backend. `cpu`, `cpu-f16`, `cpu-f32` keep its weights on the CPU backend — useful when isolating bugs to the talker vs. code-predictor or when comparing CPU and Metal end-to-end. |
 | `QWEN3_TTS_DUMP_DIR` | unset | Write per-frame intermediate tensors into the named directory. Bulky; intended for diff-harness work (`tools/dump_reference.py --backend qwen3-tts`). |
+| `QWEN3_TTS_CODEC_GPU` | unset | Route the codec decode through the main GPU scheduler instead of the CPU-only `codec_sched`. The codec is pinned to CPU by default to dodge the M1 Metal hang; on CUDA / Vulkan / etc. the hang does not apply and CPU codec is dramatically slower. On Jetson Orin AGX, codec on CPU is ~50× slower than CUDA. Distinct from `QWEN3_TTS_CODEC_FORCE_METAL`, which also moves the codec to the main GPU sched but additionally enables a per-op `ggml_backend_synchronize` trace callback for reproducing the Metal hang — useful for debugging, not for production. |
 
 ## VibeVoice — realtime streaming TTS
 
--- a/src/qwen3_tts.cpp
+++ b/src/qwen3_tts.cpp
@@ -3579,13 +3579,22 @@ static bool load_codec(qwen3_tts_context* c, const char* path) {
     hp.rms_norm_eps = core_gguf::kv_f32(meta, "qwen3tts_codec.dec.rms_norm_eps", hp.rms_norm_eps);
     core_gguf::free_metadata(meta);
 
-    // Pass 2: weights — pinned to CPU backend (see codec_sched comment in
-    // qwen3_tts_context). Override with QWEN3_TTS_CODEC_FORCE_METAL=1 to
-    // reproduce the M1 Metal crash for instrumentation.
+    // Pass 2: weights — pinned to CPU backend by default to dodge the M1
+    // Metal hang. Two override env vars route weights onto the main GPU
+    // backend instead:
+    //   QWEN3_TTS_CODEC_FORCE_METAL=1 — reproduces the M1 hang with per-op
+    //                                   tracing for instrumentation.
+    //   QWEN3_TTS_CODEC_GPU=1         — clean GPU path with no trace.
+    //                                   Use on CUDA / Vulkan where the
+    //                                   Metal hang does not apply. On
+    //                                   Jetson Orin AGX, codec on CPU is
+    //                                   ~50x slower than CUDA.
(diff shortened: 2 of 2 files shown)


CrispStrobe (approved):
LGTM — clean diff, well-justified, 40× speedup on Jetson Orin AGX is exactly the right kind of perf win for headless CUDA hosts. Default path bit-identical (`(force_metal || codec_gpu) ? c->backend : c->backend_cpu` collapses to the original `c->backend_cpu` when both env vars unset). Backwards-compatible by inspection.

The one red check (Bindings Tests / Ruby) is **pre-existing main breakage**, not introduced by this PR. Just verified by re-running the Ruby workflow on `b553546` (clean main, no PR #57 changes) and getting the identical `make: *** [Makefile:171: cmake-targets] Error 2`. Filing a separate issue to track that.

Also followed up on the side-findings you flagged in the PR description:
- **#56 (`QWEN3_TTS_O15` default-on CUDA crash)** — fixed in `61c42bf`. Default flipped back to OFF; M1 Metal users who want the speedup set `QWEN3_TTS_O15=1`. Goes back to default-ON once the CUDA `ggml_set_rows` path is verified.
- **`synthesize` ref-codes-trim cost** — worth its own issue; the cleaner fix is an internal `skip_ref_decode` knob on `qwen3_tts_synthesize` so other consumers (CLI, Python wrapper) get the same win without having to call the lower-level `synthesize_codes`+`decode_codes` pair manually.
- **`/v1/audio/speech`** — full design feedback posted on #58. Looking forward to the PR.
- **Same pattern in kokoro** — applied the `_GPU` env-var pattern to `KOKORO_GEN_FORCE_METAL` in `51ec8f0`. The new `KOKORO_GEN_GPU=1` is a clean alias for non-Mac users.

Thanks for the PR — pure C++ + ggml on Jetson is exactly the headless-deploy story this engine should support, and your benchmark + writeup quality made the review easy.

PR #63, opened 2026-05-04
feat(server,qwen3-tts): /v1/audio/speech + /v1/voices, per-request voice switch

### LINKED ISSUE #58: [Feature Request] Add /v1/audio/speech (OpenAI-compat TTS) to crispasr-server
## Summary

The HTTP server in `examples/server/server.cpp` is currently ASR-only. `docs/server.md` documents OpenAI-compat `/v1/audio/transcriptions` and `/v1/models`, but there's no equivalent for TTS — even though crispasr already ships four production TTS backends (qwen3-tts, kokoro, vibevoice, orpheus) that all expose the same `CrispasrBackend::synthesize()` interface.

I'd like to add `POST /v1/audio/speech` to the server so users can keep one TTS backend hot in memory and serve it over HTTP — the canonical "agent-friendly TTS" pattern. I've built and benchmarked a working version on Jetson Orin AGX and want to upstream it, but before opening a ~470-LOC PR I'd like your input on the design.

## What I built (working today)

A new TTS-only mode in `crispasr-server`, triggered when `--backend qwen3-tts*`. It loads the talker + codec once at startup, optionally pre-warms a default voice, and exposes:

- `POST /v1/audio/speech` — body: `{"input": "...", "voice": "<name>", "response_format": "wav"|"pcm"}`, returns `audio/wav` (24 kHz mono int16) or raw float32 PCM
- `GET /v1/voices` — list `<name>.wav` + `<name>.txt` pairs in `--voice-dir`
- `GET /v1/models` — OpenAI-compat model list (single entry, the loaded backend)
- `GET /health` — liveness

Voice profiles are filesystem-based: `<voice-dir>/<name>.wav` paired with `<name>.txt` (reference transcript). Per-request `voice` switches the active speaker prompt; the ECAPA + codec-encode cost is amortized after first use.

End-to-end measured on Jetson Orin AGX (after applying PR #57 + the `synthesize_codes` + `decode_codes` two-call pattern): RTF ~1.5 for clips from 2 s to 64 s.

Implementation lives on a fork: [`vkrmch/CrispASR feat/tts-server-endpoint`](https://github.com/vkrmch/CrispASR/tree/feat/tts-server-endpoint) — happy to share specific files or push a draft PR if it's easier to look at code than discuss in the abstract.

## Design questions

Before opening a PR, I'd appreciate input on:

1. **Single binary vs separate.** I added TTS routes inline in `crispasr-server` and dispatch on backend name early in `main()`. Alternative: a sibling `crispasr-tts-server` binary that shares helpers. Given `CrispasrBackend` already has `CAP_TTS`, I think one binary is cleaner — but happy to split if you want `server.cpp` to stay narrow.

2. **Generic over all TTS backends.** My current patch hardcodes `qwen3_tts.h` C-API calls (the qwen3-tts backend was the one I needed). The cleaner shape would route through `CrispasrBackend::synthesize()` so the same routes serve kokoro / vibevoice / orpheus too. The CLI's `whisper_params` already has the relevant TTS fields (`tts_voice`, `tts_ref_text`, `tts_codec_model`, `tts_instruct`). Worth refactoring against that interface for v1, or land qwen3-tts-only and generalise in a follow-up?

3. **Voice management.** Filesystem-based `<voice-dir>/<name>.{wav,txt}` is the path of least resistance and matches how voice GGUFs already get distributed. Future options: `POST /v1/voices/upload` (multipart) to add at runtime, or inline `ref_audio_b64` in `/v1/audio/speech` requests. Land now or punt?

4. **Streaming.** OpenAI's `/v1/audio/speech` supports chunked-transfer streaming. With cpp-httplib it's achievable; perceived latency would drop substantially since playback can start before AR finishes. Out of scope for v1 or worth landing together?

5. **Output format.** I currently support `wav` (16-bit PCM int16, 24 kHz mono) and `pcm` (raw float32 native qwen3-tts output). Adding mp3/opus would need an embedded encoder or shelling out to ffmpeg. Land as wav-only and let clients transcode, or include format conversion?

6. **Auth.** Existing `/inference` has no auth. `docs/server.md` mentions `CRISPASR_API_KEYS` but I don't see it wired in `server.cpp` — happy to land that bit if it's been on your list, otherwise I'll mirror the existing no-auth shape.

Happy to PR whatever shape you want — or if you'd prefer to design from scratch, I'll scrap mine and follow your direction.

## Related
- #57 — `QWEN3_TTS_CODEC_GPU` env var (perf prerequisite for any usable codec-on-GPU TTS server)
- #56 — `QWEN3_TTS_O15` default-on CUDA crash (workaround: set `QWEN3_TTS_O15=0`)

Thanks for the project — the C++ + ggml runtime is a great fit for headless Jetson deployments.

vkrmch:
Closes #58.

Adds the OpenAI-compatible TTS endpoint to the persistent `crispasr-server`, following the design we converged on in #58. Hands the request through the existing `CrispasrBackend::synthesize()` abstraction so kokoro / qwen3-tts / vibevoice / orpheus all get the routes for free as soon as the user loads them.

## Routes

```
POST /v1/audio/speech
  Body: application/json
    {
      "input":           "<text to synthesize>",       (required)
      "voice":           "<name in --voice-dir>",      (optional)
      "response_format": "wav" | "pcm"                 (optional, default "wav")
    }
  Returns:
    200 audio/wav                  16-bit PCM int16 WAV at 24 kHz mono
    200 application/octet-stream   raw float32 PCM (response_format=pcm)
    400  loaded backend lacks CAP_TTS, missing/empty input, malformed body,
         or `voice` provided but server has no --voice-dir configured
    422  voice name not found in --voice-dir
    500  backend->synthesize returned empty audio
    503  model still loading

GET /v1/voices
  Returns: { "voices": [ { "name": "<stem>", "format": "wav"|"gguf" }, ... ] }
  Source: directory listing of --voice-dir for *.{wav,gguf} files
```

## Design choices (per #58 thread)

- **Single binary, capability-gated.** Routes register on every backend; the TTS handlers reject with 400 when `backend->capabilities() & CAP_TTS` is unset. No backend-name string match — kokoro/qwen3-tts/vibevoice/orpheus get the routes automatically once they advertise `CAP_TTS`.
- **Generic via the abstraction.** Calls `backend->synthesize(text, params)` and writes `vector<float>` 24 kHz mono PCM to either a WAV (int16 RIFF) or raw float32. No backend-specific code in the route handler — the server doesn't include `qwen3_tts.h`.
- **Voice management is filesystem.** `<voice-dir>/<name>.wav` paired with `<name>.txt` (Qwen3-TTS ICL ref-text), or `<voice-dir>/<name>.gguf` (baked voice pack). Maps the request's `voice` field directly to `whisper_params::tts_voice`. Upload endpoint and inline `ref_audio_b64` deferred per #58.
- **WAV + PCM only.** `pcm` is free since the backend already returns `vector<float>`. `mp3` / `opus` deferred.
- **No streaming.** Deferred per #58.
- **No auth changes.** Reuses the existing `is_authorized` / `require_auth` shape from the file. `CRISPASR_API_KEYS` already covers the new routes since they go through `require_auth`.

## Per-request voice switching

The qwen3-tts adapter (`crispasr_backend_qwen3_tts.cpp`) previously cached voice-load behind a single boolean (`voice_loaded_`) that ignored later changes to `params.tts_voice` / `params.tts_ref_text` / `params.tts_instruct`. The CLI's single-shot use case never noticed because each invocation creates a fresh process; the server caller hits it the first time it tries to switch voice.

Replaced with a composite voice-key cache (`last_voice_key_`):
```cpp
// VoiceDesign:    "vd:" + tts_instruct
// CustomVoice:    "cv:" + tts_voice
// Base (WAV/pack): "base:" + tts_voice + "\x01" + tts_ref_text
```
Re-runs `qwen3_tts_set_voice_prompt_with_text` / `qwen3_tts_set_speaker_by_name` / `qwen3_tts_set_instruct` only when the key actually changes. CLI single-shot behaviour is unchanged (first call still loads once).

## CMake

Added `json_cpp` (the existing `examples/CMakeLists.txt` INTERFACE library that exposes `examples/json.hpp`) to `crispasr-cli`'s link list — needed for `nlohmann::json::parse(req.body)`. No new dependencies.

## Smoke test (Jetson Orin AGX)

```
$ crispasr --server --backend qwen3-tts-1.7b-base \
    -m ~/.cache/crispasr/qwen3-tts-12hz-1.7b-base-q8_0.gguf \
    --codec-model ~/.cache/crispasr/qwen3-tts-tokenizer-12hz.gguf \
    --voice-dir ~/voice-ref --host 0.0.0.0 --port 11441
crispasr-server: backend 'qwen3-tts-1.7b-base' loaded, model '...12hz-1.7b-base-q8_0.gguf'
crispasr-server: listening on 0.0.0.0:11441
  POST /inference                  — upload audio (native JSON)
  POST /v1/audio/transcriptions    — OpenAI-compatible API
  POST /v1/audio/speech            — TTS (OpenAI-compatible)
  POST /load                       — hot-swap model
  GET  /health                     — server status
  GET  /backends                   — list backends
  GET  /v1/models                  — model info
  GET  /v1/voices                  — list voices in --voice-dir

$ curl -s http://nova:11441/v1/voices
{"voices": [{"name": "vik", "format": "wav"}, ...]}

$ curl -s -X POST http://nova:11441/v1/audio/speech \
    -H "Content-Type: application/json" \
    -d '{"input": "Hello there.", "voice": "vik"}' \
    -o /tmp/out.wav -w "%{http_code}\n"
200
crispasr-server: synthesized 3.8s audio in 25.50s (RTF=6.78) voice='vik' format=wav
```

Output is a valid 16-bit PCM 24 kHz mono WAV that plays back the input cloned in the reference voice.

The 6.78 RTF in this trace is from `qwen3_tts_synthesize` re-decoding the ref codes through the codec on every call (~16s of constant cost on Orin AGX from the 26-second reference). That's worth fixing internally as a `skip_ref_decode` knob on `qwen3_tts_synthesize` — I'll file a separate issue for it as agreed in #58.

## Out of scope (deferred per #58, happy to pick up in follow-ups)

- `POST /v1/voices` (multipart upload) and inline `ref_audio_b64` for runtime voice provisioning
- mp3 / opus encoding
- Streaming response (chunked transfer)
- `CRISPASR_API_KEYS` audit across `/inference` + new TTS routes (already wired via `require_auth`, but worth a follow-up review)
- `skip_ref_decode` knob on `qwen3_tts_synthesize` to drop the constant ref-decode cost (#58 thread)

Thanks for the design feedback on #58 — converging on the abstraction-friendly shape made the implementation clean.

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -406,6 +406,11 @@ static bool whisper_params_parse(int argc, char** argv, whisper_params& params) 
             params.tts_ref_text = ARGV_NEXT;
         } else if (arg == "--instruct") {
             params.tts_instruct = ARGV_NEXT;
+        } else if (arg == "--voice-dir") {
+            // Server mode: directory of <name>.wav (+ <name>.txt) or
+            // <name>.gguf voice profiles. Used by POST /v1/audio/speech
+            // to resolve the request's "voice" field.
+            params.tts_voice_dir = ARGV_NEXT;
         } else if (arg == "--tts-trim-silence") {
             params.tts_trim_silence = true;
         } else if (arg == "--text") {
@@ -688,6 +693,8 @@ static void whisper_print_usage(int /*argc*/, char** argv, const whisper_params&
     fprintf(stderr, "             --instruct \"TEXT\"        natural-language voice description "
                     "(qwen3-tts VoiceDesign only; replaces --voice)\n");
     fprintf(stderr, "             --codec-model FNAME      qwen3-tts codec GGUF (defaults to sibling of -m)\n");
+    fprintf(stderr, "             --voice-dir PATH         server: dir of <name>.wav (+ <name>.txt) or "
+                    "<name>.gguf voice profiles for POST /v1/audio/speech\n");
     fprintf(stderr, "             --tts-steps N            [%-7d] DPM-Solver++ steps (10-20, vibevoice only)\n",
             params.tts_steps);
     fprintf(stderr, "             --tts-trim-silence       [%-7s] trim leading silence from TTS output\n",
--- a/examples/cli/crispasr_backend_qwen3_tts.cpp
+++ b/examples/cli/crispasr_backend_qwen3_tts.cpp
@@ -109,14 +109,28 @@ public:
         if (!ctx_ || text.empty())
             return {};
 
-        // Voice prompt: load once on first synthesise call. Four paths:
+        // Voice prompt: re-load only when the requested identity changes.
+        // Four mutually-exclusive paths (gated by the loaded model variant):
         //   --voice <name>       → CustomVoice fixed-speaker selection
         //                          (only when the loaded model is CustomVoice)
         //   --voice X.gguf       → baked voice pack (Base)
         //   --voice X.wav --ref-text "..." → runtime ECAPA + codec encoder (Base)
         //   --instruct "..."     → VoiceDesign natural-language description
         //                          (only when the loaded model is VoiceDesign)
-        if (!voice_loaded_) {
+        //
+        // Cache the composite identity in `last_voice_key_` so the CLI's
+        // single-shot use case still pays the load cost only once, while
+        // server-mode callers can switch voice per request just by changing
+        // `params.tts_voice` (or `params.tts_ref_text` / `params.tts_instruct`).
+        std::string voice_key;
+        if (qwen3_tts_is_voice_design(ctx_)) {
+            voice_key = "vd:" + params.tts_instruct;
+        } else if (qwen3_tts_is_custom_voice(ctx_)) {
+            voice_key = "cv:" + params.tts_voice;
+        } else {
+            voice_key = "base:" + params.tts_voice + "\x01" + params.tts_ref_text;
+        }
+        if (voice_key != last_voice_key_) {
             if (qwen3_tts_is_voice_design(ctx_)) {
                 // VoiceDesign: --instruct is required, --voice has no role.
                 if (!params.tts_voice.empty() && !params.no_prints) {
@@ -183,7 +197,7 @@ public:
                     }
                 }
             }
-            voice_loaded_ = true;
+            last_voice_key_ = voice_key;
         }
 
         int n = 0;
@@ -200,12 +214,12 @@ public:
             qwen3_tts_free(ctx_);
             ctx_ = nullptr;
         }
-        voice_loaded_ = false;
+        last_voice_key_.clear();
     }
 
 private:
     qwen3_tts_context* ctx_ = nullptr;
-    bool voice_loaded_ = false;
+    std::string last_voice_key_;
 };
 
 } // namespace
--- a/examples/cli/crispasr_server.cpp
+++ b/examples/cli/crispasr_server.cpp
@@ -26,7 +26,7 @@
 
 #include "common-crispasr.h" // read_audio_data
 #include "../server/httplib.h"
-#include "../server/json.hpp"
+#include "../json.hpp"
 
 #include <atomic>
 #include <cerrno>
--- a/examples/cli/crispasr_server.cpp
+++ b/examples/cli/crispasr_server.cpp
@@ -9,10 +9,12 @@
 // Endpoints:
 //   POST /inference                   — transcribe (native JSON)
 //   POST /v1/audio/transcriptions     — OpenAI-compatible endpoint
+//   POST /v1/audio/speech             — TTS (OpenAI-compatible; CAP_TTS only)
 //   POST /load                        — hot-swap model
 //   GET  /health                      — server status
 //   GET  /backends                    — list available backends
 //   GET  /v1/models                   — OpenAI-compatible model list
+//   GET  /v1/voices                   — list voices in --voice-dir (CAP_TTS only)
 //
 // Adapted from examples/server/server.cpp for multi-backend support.
 
@@ -24,13 +26,17 @@
 
 #include "common-crispasr.h" // read_audio_data
 #include "../server/httplib.h"
+#include "../server/json.hpp"
 
 #include <atomic>
 #include <cerrno>
 #include <chrono>
+#include <cmath>
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <filesystem>
 #include <fstream>
 #include <memory>
 #include <mutex>
@@ -276,6 +282,100 @@ static transcription_result do_transcribe(const httplib::MultipartFormData& audi
 
     result.ok = true;
     return result;
+}
+
+// ---------------------------------------------------------------------------
+// TTS helpers
+// ---------------------------------------------------------------------------
+
+// Read a whole file into a string. Returns empty string on failure.
+static std::string read_text_file(const std::string& path) {
+    std::ifstream f(path);
+    if (!f.good())
+        return "";
+    std::stringstream ss;
+    ss << f.rdbuf();
+    return ss.str();
+}
+
+// Resolve a voice request name to an absolute path (and optional ref-text)
+// from the configured voice-dir. Looks for <name>.wav (paired with
+// <name>.txt for the reference transcription used by Qwen3-TTS ICL
+// prefill) or <name>.gguf (baked voice pack). Returns true on hit;
+// sets out_path and out_ref_text.
+static bool resolve_voice(const std::string& voice_dir, const std::string& name, std::string& out_path,
+                          std::string& out_ref_text) {
+    if (voice_dir.empty() || name.empty())
+        return false;
+    const std::string wav = voice_dir + "/" + name + ".wav";
+    const std::string txt = voice_dir + "/" + name + ".txt";
+    const std::string gguf = voice_dir + "/" + name + ".gguf";
+    {
+        std::ifstream f(wav);
+        if (f.good()) {
+            out_path = wav;
+            out_ref_text = read_text_file(txt);
+            return true;
+        }
+    }
+    {
+        std::ifstream f(gguf);
+        if (f.good()) {
+            out_path = gguf;
+            out_ref_text.clear();
+            return true;
+        }
+    }
+    return false;
+}
+
+// Build a 16-bit PCM RIFF WAV from float32 samples in [-1, 1].
+// Mono, sample_rate Hz. Header is the standard 44-byte PCM-fmt RIFF.
+// Returned buffer is contiguous and safe to pass to res.set_content().
+static std::string make_wav_int16(const float* pcm, int n_samples, int sample_rate) {
+    const uint16_t num_channels = 1;
+    const uint16_t bits_per_sample = 16;
+    const uint32_t byte_rate = sample_rate * num_channels * (bits_per_sample / 8);
+    const uint16_t block_align = num_channels * (bits_per_sample / 8);
+    const uint32_t data_size = (uint32_t)n_samples * block_align;
+    const uint32_t riff_size = 36 + data_size;
+
+    std::string out;
+    out.reserve(44 + (size_t)data_size);
+    auto put_u32 = [&](uint32_t v) {
+        out.push_back((char)(v & 0xff));
+        out.push_back((char)((v >> 8) & 0xff));
+        out.push_back((char)((v >> 16) & 0xff));
+        out.push_back((char)((v >> 24) & 0xff));
+    };
+    auto put_u16 = [&](uint16_t v) {
+        out.push_back((char)(v & 0xff));
+        out.push_back((char)((v >> 8) & 0xff));
+    };
+    out.append("RIFF", 4);
+    put_u32(riff_size);
+    out.append("WAVE", 4);
+    out.append("fmt ", 4);
+    put_u32(16); // PCM fmt chunk size
+    put_u16(1);  // PCM format
+    put_u16(num_channels);
+    put_u32(sample_rate);
+    put_u32(byte_rate);
+    put_u16(block_align);
+    put_u16(bits_per_sample);
+    out.append("data", 4);
+    put_u32(data_size);
+    out.resize(out.size() + (size_t)data_size);
+    int16_t* dst = reinterpret_cast<int16_t*>(&out[out.size() - data_size]);
+    for (int i = 0; i < n_samples; i++) {
+        float s = pcm[i];
+        if (s > 1.0f)
+            s = 1.0f;
+        if (s < -1.0f)
+            s = -1.0f;
+        dst[i] = (int16_t)std::lround(s * 32767.0f);
+    }
+    return out;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


CrispStrobe (changes requested):
Thanks for the careful design — the capability-gated dispatch and the `last_voice_key_` rework are clean. Tested locally on macOS Metal: applying the diff and running `cmake --build build-ninja-compile --target crispasr-cli` fails to compile out of the box, so I'm flagging blocking items before the cosmetic ones.

## Blocking

### 1. Build break — `#include "../server/json.hpp"` doesn't exist

`examples/cli/crispasr_server.cpp:29` includes `"../server/json.hpp"`, but the only nlohmann header in the tree is `examples/json.hpp`. There is no `examples/server/json.hpp` (never has been — `git log --all --diff-filter=D` returns nothing). The PR also doesn't modify `examples/cli/CMakeLists.txt`, contradicting the description's claim that `json_cpp` was added to `crispasr-cli`'s link list.

Verified by checkout-and-build:

```
/Users/.../examples/cli/crispasr_server.cpp:29:10: fatal error: '../server/json.hpp' file not found
   29 | #include "../server/json.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Minimal fix — change to `#include "../json.hpp"`. `examples/` is already in the include path via `target_include_directories(${TARGET} PRIVATE ${CMAKE_SOURCE_DIR}/examples)` (transitively, via the existing `examples/server` include and the global examples include). Confirmed builds cleanly with that one-character fix on Metal.

If you actually want the `json_cpp` INTERFACE library route (cleaner — keeps the include explicit), then `examples/cli/CMakeLists.txt`'s `target_link_libraries(${TARGET} PRIVATE ...)` needs `json_cpp` added, and the include can stay as `\"json.hpp\"` (no relative path).

### 2. Path traversal in `resolve_voice`

`examples/cli/crispasr_server.cpp:308-310`:

```cpp
const std::string wav  = voice_dir + "/" + name + ".wav";
const std::string txt  = voice_dir + "/" + name + ".txt";
const std::string gguf = voice_dir + "/" + name + ".gguf";
```

`name` comes straight from the request body. With `voice = \"../../etc/ssh/ssh_host_rsa_key\"` the server tries to open `<voice-dir>/../../etc/ssh/ssh_host_rsa_key.wav`. The forced `.wav` / `.gguf` suffix narrows the attack surface (can only read files that happen to end in those extensions) but still leaks any sibling `.wav` / `.gguf` on the box — including the loaded model GGUF. Also, the `<name>.txt` paired-read happens *after* the `.wav` succeeds, which means an attacker who can drop a sibling `.wav` next to a sensitive `.txt` (e.g. a config file under `--voice-dir/..`) gets that text read into `tts_ref_text` and echoed back through synthesis loss patterns.

Fix: reject `name` containing `/`, `\`, NUL, or `..` segments, or — more robust — resolve the path and verify `lexically_normal()` is contained by `voice_dir`'s normalized form:

```cpp
namespace fs = std::filesystem;
const fs::path base = fs::weakly_canonical(voice_dir);
const fs::path target = fs::weakly_canonical(base / (name + \".wav\"));
auto rel = target.lexically_relative(base);
if (rel.empty() || rel.native()[0] == '.') return false; // escape
```

## Should fix before merge

### 3. `/v1/voices` doesn't gate on CAP_TTS despite the file-header doc claiming it does

The route comment in the changed banner says `GET /v1/voices — list voices in --voice-dir (CAP_TTS only)`. The handler at `examples/cli/crispasr_server.cpp:780+` only checks `require_auth` — an ASR-only backend gets a `200 {\"voices\": []}` instead of a 4xx. Pick one: either add `if (!(backend->capabilities() & CAP_TTS)) return json_error(res, 400, ...)`, or drop the \"CAP_TTS only\" claim from the docs. The startup banner already gates the `/v1/voices` line on CAP_TTS — the route handler should match.

### 4. Per-request voice swap is silently a no-op for chatterbox

The PR's claim is that all CAP_TTS backends get the route \"for free\". True for kokoro / qwen3-tts / vibevoice / orpheus — they all read `params` in `synthesize()`. But `examples/cli/crispasr_backend_chatterbox.cpp` has:

```cpp
std::vector<float> synthesize(const std::string& text, const whisper_params& /*params*/) override {
```

— `params` is unused, so `voice` in the request body is silently ignored when a chatterbox model is loaded. Either wire chatterbox to consume `params.tts_voice` or call this caveat out in the route docstring (\"voice routing is per-backend; chatterbox currently ignores the request voice\"). Not blocking but will cause confused bug reports.

## Worth knowing, not blocking

- **Synthesis holds `model_mutex` for the entire RTF-6.78 duration.** A 25-second synth blocks `/inference` and `/v1/audio/transcriptions` on the same server. Matches the ASR pattern, so it's not a regression — but the existing ASR side is fast (~0.5s for short clips); TTS makes the contention much more visible. Worth a note in the route docstring, or a follow-up to allow read-only routes (`/v1/voices`, `/v1/models`, `/health`, `/backends`) to bypass the model mutex.
- **422 vs 404 for unknown voice** — OpenAI conventions point at 404 for \"resource not found\". 422 is defensible as \"semantically invalid input\" but a stock OpenAI client may not handle it the way you'd expect. Either is fine; flag for awareness.
- **Voice precedence** — `<name>.wav` wins over `<name>.gguf` if both exist. Worth a sentence in the `resolve_voice` doc comment so users with both don't get surprised.
- **Tests** — there are none; a `tests/server_tts_smoke.sh` (curl-based, just hitting `/v1/voices` and posting a tiny synth then validating the WAV header) would be cheap and would catch the include-path regression that this review caught manually.

## Things I liked

- `last_voice_key_` is a real correctness fix that the CLI also benefits from (the old `voice_loaded_ = true` would have ignored a future `--voice` change in any long-lived caller, not just the server). Preserving the load-once behaviour for single-shot CLI by virtue of key equality is the right shape.
- Capability-gated dispatch instead of backend-name string matching is exactly the design that makes adding a 6th TTS backend zero work in the route handler.
- WAV serializer is straightforward — RIFF size formula is right, sample clamp+round is right, no obvious endianness slip on little-endian targets (you write byte-by-byte, so it's correct on both ends — though the codebase only ships LE platforms anyway).
- `request_format=pcm` is a nice addition; saves a serialization roundtrip for callers that want to feed the float32 into another DSP stage.

Happy to land a small follow-up commit with the include fix + the path-traversal sanitization once you've reviewed the design points above. The qwen3-tts and route logic is otherwise solid.

CrispStrobe:
Following up on the line-level review with the conceptual / architectural angle, reading the PR against the design we converged on in #58 and against the existing C-ABI + adapter layering. Splitting this from the line review because it's about *where* code lives rather than *what* the code does.

## Architecturally sound at the macro level

- **Capability-gated dispatch via `backend->capabilities() & CAP_TTS`** is exactly what we agreed in #58 ("not the backend name string"). The route handler stays at the C++ adapter (`CrispasrBackend`) seam — zero `qwen3_tts.h` / `kokoro.h` includes leak in. New TTS backends (when CAP_TTS is added) get the route automatically.
- **`std::lock_guard<std::mutex>` around `backend->synthesize()`** matches the per-context single-thread invariant the C-ABI imposes. Each `qwen3_tts_context*` / `kokoro_context*` / etc. owns mutable state (KV cache, voice-prompt cache); concurrent calls would corrupt it. Mirroring the ASR-side mutex is the right shape.
- **Polymorphic `synthesize(text, params)`** with a pass-through `whisper_params` is the same dispatch shape `/inference` and `/v1/audio/transcriptions` use. No new abstraction was needed; the design just exercises what was already there.
- **The `last_voice_key_` rework is a correctness fix**, not just a feature. The old `voice_loaded_` boolean would silently ignore any future change to `params.tts_voice` after first call — which would have bitten any long-lived caller (Python wrapper, future SDK). The server happens to be the first such caller; the CLI single-shot keeps working because key equality still loads-once.

## One substantive divergence from #58

#58's design comment included:

> *\"The `voice` request field maps directly to `whisper_params::tts_voice` — backend decides whether to interpret it as a filename, a baked-pack path, or a preset name (orpheus's `tara`/`leah`).\"*

PR #63 instead does **server-side filesystem resolution** in `resolve_voice()` and passes the resolved *path* into `params.tts_voice`. That moves the voice-vocabulary decision from the backend adapter into the route handler. Concrete consequences:

1. **Breaks qwen3-tts CustomVoice models silently.** `crispasr_backend_qwen3_tts.cpp:147` checks `ends_with_ci(spk_name, \".wav\")` and falls back to \"first speaker\" with a stderr warning. So a request `{\"voice\": \"vivian\"}` against a CustomVoice model becomes: server resolves to `<voice-dir>/vivian.wav` → adapter sees it ends in `.wav` → ignores it → wrong voice synthesized, **no error returned to the client**. The user gets a 200 + audio, but it's not the voice they asked for.
2. **Breaks orpheus presets** like `tara`/`leah` that the #58 thread specifically called out. The server tries `<voice-dir>/tara.wav`, fails → 422, even though orpheus's adapter could handle the preset name natively if the string were passed through.
3. **Adds the path-traversal attack surface** flagged in the line review. Pass-through wouldn't have it — the backend's name-to-voice lookup is a registry, not the filesystem.
4. **`/v1/voices` listing reflects the filesystem, not the backend's actual voice set.** A CustomVoice user sees voices that aren't usable; doesn't see voices that are.

The shape that matches the agreed design:

```cpp
// Route handler — no FS resolution, no path manipulation
whisper_params rp = params;
rp.tts_voice = body.value(\"voice\", rp.tts_voice);  // pass-through
// (rp.tts_voice_dir already populated from server flag)
auto pcm = backend->synthesize(text, rp);
```

`--voice-dir` becomes a *convenience layer*: `/v1/voices` lists what's there, and the backend adapter can opt to look in `params.tts_voice_dir` if its own name resolution misses. Only qwen3-tts (Base variant) needs this, and its adapter is already path-aware. The other adapters keep doing whatever they do today — the route handler doesn't dictate.

That shape also makes the route handler dramatically simpler: no `resolve_voice`, no `read_text_file`, no path arithmetic, no traversal-sanitization burden. The 422 response goes away (the backend already returns empty PCM → 500 if a name doesn't resolve, or the adapter can be taught to return a richer error code in a follow-up).

## One OpenAI-spec mismatch

`response_format: \"pcm\"` returning **float32**. OpenAI's `pcm` is **24 kHz signed 16-bit LE mono**. A stock OpenAI client passing `\"pcm\"` will interpret the float32 bytes as int16 and get garbage. The #58 thread agreed on \"raw float32 PCM (response_format=pcm)\" — so this matches what we discussed, but it'll bite real OpenAI-spec clients. Two clean fixes: rename to `\"f32\"` (and reject `\"pcm\"` with a 400 pointing at OpenAI's spec), or actually emit int16. I'd lean toward `f32` since the float path is genuinely useful for downstream DSP.

## Comparison with the closed predecessor (PR #4)

PR #4 was the community attempt at `/v1/audio/transcriptions`; it got closed and the maintainer wrote the route from scratch. PR #63 has the same trajectory potential — the macro design is right, the surface details don't quite match the codebase's existing idiom. The fix is small enough that landing it from this PR is the right call (versus closing + rewriting): the diff is mostly things you'd write the same way.

## What I'd do differently if writing it fresh

Nothing structural. The capability gating, the polymorphic dispatch, the mutex placement, the WAV serializer, the `--voice-dir` flag — all the right calls. The two changes I'd make: move voice resolution back to the backend adapters (per #58), and either rename `pcm` to `f32` or actually emit int16. Both are local fixes inside the route handler, no architectural rework.

CrispStrobe:
Thanks again for the design work + the smoke run on Orin AGX — landed yours as `cd30c46` (squash kept your authorship through the include-fix commit). Following up on the review notes with what shipped on top this session:

**Corrective batch** — addresses the design-review concerns from the two review comments above:

* `d35940b` — voice-resolution shape: per the #58 design comment ("`voice` request field maps directly to `whisper_params::tts_voice` — backend decides"), removed `resolve_voice` from the route handler and taught the qwen3-tts adapter to do bare-name resolution against `--voice-dir` itself. Restores CustomVoice + orpheus presets, eliminates the path-traversal surface (sanitization at the adapter, before any FS call), `/v1/voices` behaviour unchanged.
* `c317771` — `/v1/voices` CAP_TTS gate (matches the startup banner).
* `b263566` — `pcm` → `f32` rename (interim — see below for the real `pcm`).
* `15c0fd4` — pre-existing bug surfaced by the smoke test: `set_error_handler` was clobbering every 4xx body, including ours (and `/inference`'s). Guarded on `res.body.empty()`.

**OpenAI parity round 1** — gap analysis written up as PLAN §75:

* `b932fa9` — §75a + §75b: `model` field accepted-and-logged, `--tts-max-input-chars N` cap (default 4096), `instructions` mapped to `params.tts_instruct` (silently ignored on non-VoiceDesign so OpenAI clients don't 4xx), real OpenAI `pcm` (24 kHz signed 16-bit LE, no header — `f32` stays as the crispasr extension), `speed` 0.25..4.0 via post-synth linear resampler, `--cors-origin '*'|<origin>` opt-in CORS with preflight, OpenAI error shape (`code` / `param` fields).
* `0bff2d7` — §75d: your #66 sentence-chunking proposal upstreamed as a standalone `examples/cli/crispasr_tts_chunking.{h,cpp}` helper with two extensions over the prototype: CJK ideographic full stop (U+3002) + Devanagari danda (U+0964) recognised as primary terminators, and decimal-aware (`1.5 today` doesn't split). Default-on, no opt-in flag, route handler iterates + concatenates with 200 ms silence pad. Synth log line now reports `chunks=N`. See [issue #66 comment](https://github.com/CrispStrobe/CrispASR/issues/66) for details.

**Test coverage added:**
* `tests/test-server-tts.sh` — 35 integration smoke assertions (CORS preflight + headers, all error paths with code/param, real `pcm` body shape, `speed=2.0` produces ~half samples, `model`/`instructions` accepted, voice switching across two CustomVoice speakers, multi-sentence chunking).
* `tests/test_server_wav_writer.cpp` — 75 Catch2 assertions across 15 cases for the WAV + PCM serializers (RIFF byte layout, clamp/round, byte-identical bodies after header strip).
* `tests/test_server_chunking.cpp` — 76 Catch2 assertions across 19 cases (empty/whitespace, every terminator script, decimal preservation, abbreviation over-split documented, max_chars fallback, pathological no-whitespace blob, concat edge cases).

186 total green against qwen3-tts-customvoice 0.6B Q8_0 on M1 Metal.

The qwen3-tts `last_voice_key_` rework you did was a real correctness fix that the CLI also benefits from — the chunking path leans on it heavily because it amortises the per-call voice-load cost across the chunk loop. Solid contribution.

Docs at `docs/server.md` (TTS endpoint section), `docs/tts.md` (HTTP pointer), `README.md` (quick-start TTS curl). Plan in `PLAN.md` §75 with the deferred items (native-backend duration knobs, streaming response, mp3/opus, voice upload).

Issue #74, opened 2026-05-07
[bug] vibevoice-tts 1.5B WAV-clone path produces generic voice (doesn't actually clone the reference speaker)


vkrmch:
## Summary

The `vibevoice-tts` 1.5B WAV-cloning path through `VIBEVOICE_VOICE_AUDIO` produces clear, intelligible audio but **doesn't reproduce the reference speaker's voice** — output sounds like a generic voice from the model's prior, not the WAV reference. Tested across Q4_K and Q8_0 quantizations and with/without -25 dB FS pre-normalization on a 26-second 24 kHz mono reference. Repro is consistent: the listener can confirm the synthesis worked but cannot identify the reference speaker.

Filing this as a quality bug because the workflow is documented (README + HF discussion #51 mention WAV cloning for the 1.5B GGUF) but the resulting output isn't actually cloned. Two adjacent surface-level findings that compound the problem are listed at the bottom.

## Environment

- Jetson Orin AGX 64 GB, CUDA 12.8, sm_87
- CrispASR HEAD `a471a3c` (current main + my [feat/vibevoice-voice-dir-resolution](https://github.com/vkrmch/CrispASR/tree/feat/vibevoice-voice-dir-resolution) adapter — the bug reproduces with or without that branch)
- Models tested: `cstr/vibevoice-1.5b-tts-q4_k.gguf` (1.71 GB) and `cstr/vibevoice-1.5b-tts-q8_0.gguf` (3.0 GB)
- Reference WAV: 26.7 s English speech, 24 kHz mono int16

## Reproducer

```bash
# Place the 1.5B GGUF in a directory with no sibling vibevoice-voice-*.gguf
# (otherwise the adapter auto-loads emma instead of taking the env-var path).
mkdir -p ~/.cache/crispasr-1.5b
mv ~/.cache/crispasr/vibevoice-1.5b-tts-q8_0.gguf ~/.cache/crispasr-1.5b/

env VIBEVOICE_VOICE_AUDIO=~/voice-ref/vik.wav \
    ./build/bin/crispasr \
    --backend vibevoice-tts \
    -m ~/.cache/crispasr-1.5b/vibevoice-1.5b-tts-q8_0.gguf \
    --tts "Hello, this is a voice clone test." \
    --tts-output /tmp/out.wav
```

The log confirms the WAV is encoded:

```
encoded voice reference: 201 frames from /home/vik/voice-ref/vik.wav
prompt 239 tokens (201 voice frames)
```

[...]

PR #78, opened 2026-05-07
Fix vibevoice 1.5B tts voice cloning, CFG, and WAV parsing (Addresses #74)

DBMePls:
Hey @CrispStrobe,

Full disclosure: I've been vibecoding this with an LLM, but I managed to fix the voice cloning problem. 

To test it, I resampled the jfk clip from the samples folder with ffmpeg to 24kHz and ran it:
```bash
ffmpeg -i samples/jfk.wav -ar 24000 jfk_24k.wav
./build/bin/crispasr --tts "Hello, how are you today? Good? Me too. It has been a long day." -m vibevoice-1.5b-tts-q4_k.gguf --tts-output output.wav --voice jfk_24k.wav
```
It produced a cloned voice and it seems to be working.

Here's what I changed:

- The reference WAV was being routed through `vibevoice_encode_speech()`, which runs it through *both* the acoustic and semantic encoders. Voice cloning only wants acoustic latents. Ripped that out, routed it through just the acoustic encoder, and applied the scaling/bias factors before the connector. The cloned voice actually sounds like the speaker now.
- The tts model's negative condition was just a dummy array of zeros. Swapped this to extract the actual hidden state of the `<|vision_start|>` token (`run_qwen2_prefill_no_kv`). Way better conditioning now. Sounds more like original voice.
- The blind 44-byte `fseek` was breaking/garbling audio if the WAV had FFmpeg metadata tags or non-standard headers. Wrote a quick RIFF scanner to actually hunt for the `"data"` chunk. (Also fixed the float conversion divisor from 32767 to 32768).
- The recent -25 dB FS normalization was hard-clamping at 1.0/-1.0, which caused audible clipping if the RMS scaling pushed peaks too high. Changed it to do a uniform soft-scale down if the max peak exceeds 1.0.
- Tossed in a quick pure-inference RTF calculation at the end (wrapped behind your existing `VIBEVOICE_BENCH` flag so it doesn't spam).

### DIFF
--- a/examples/cli/crispasr_backend_vibevoice.cpp
+++ b/examples/cli/crispasr_backend_vibevoice.cpp
@@ -128,8 +128,14 @@ public:
                 return {};
             }
             const std::string gguf_path = params.tts_voice_dir + "/" + voice_path + ".gguf";
-            if (file_exists(gguf_path))
+            const std::string wav_path = params.tts_voice_dir + "/" + voice_path + ".wav";
+            if (file_exists(gguf_path)) {
                 voice_path = gguf_path;
+            } else if (file_exists(wav_path)) {
+                voice_path = wav_path;
+            } else {
+                fprintf(stderr, "crispasr[vibevoice-tts]: warning: neither '%s' nor '%s' were found on disk!\n", gguf_path.c_str(), wav_path.c_str());
+            }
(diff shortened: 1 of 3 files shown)


CrispStrobe:
thanks. looks good at first glance but will look into this properly over the weekend.

PR #90, opened 2026-05-14
Skip finalized stream VAD slices before decode

CKwasd:
## Summary

Move finalized/straddling VAD slice filtering ahead of `backend->transcribe()` in JSON streaming mode.

Previously, the stream loop decoded each VAD slice first, then the JSON state machine skipped fully finalized slices or re-decoded the post-finalized tail for straddling slices. With small `--stream-step` values, rolling-window audio could repeatedly decode already-finalized speech until it aged out of `--stream-length`.

This change avoids that wasted backend work by deciding whether a VAD slice is old, straddling, or new before calling `backend->transcribe()`.

## Details

- Skip fully finalized VAD slices before backend decode.
- Decode only the post-finalized tail for straddling slices.
- Preserve the existing 2s minimum tail gate.
- Leave non-JSON streaming behavior unchanged.

## Manual Test Notes

Tested JSON streaming with:

```text
--backend cohere
-l ja
--stream-json
--stream-final-mode redecode
--stream-final-on-silence-ms 300
--stream-length 8000
--stream-step 500
--vad
-vt 0.65
-vspd 200
-vsd 300
-vp 100
```

`dbg_lag_sec` includes an initial wall-clock offset from stream setup, so the useful signal is whether it keeps increasing over time.

In this A/B run, normalized peak backlog growth dropped from about `+13.03s` to `+4.34s`, a reduction of roughly `67%`.

```text
1 - 4.34 / 13.03 = 0.667 ~= 67%
```

Before this change, lag grew substantially during the run before later silence allowed the stream to catch up:

| Point | dbg_lag_sec | Growth from baseline |
|---|---:|---:|
| first post-track event, t=1.0s | ~9.75s | 0.00s |
| t=14.98s | ~14.25s | +4.50s |
| t=29.98s | ~15.95s | +6.20s |
| t=45.0s | ~19.72s | +9.97s |
| t=57-58s | ~22.78s | +13.03s |
| t=75.5s, after catch-up silence | ~16.36s | +6.61s |

After this change, lag stayed roughly bounded around the initial offset:

| Point | dbg_lag_sec | Growth from baseline |
|---|---:|---:|
| first post-track event, t=1.0s | ~24.46s | 0.00s |
| t=15.5s | ~25.04s | +0.58s |
| t=29.5s | ~26.30s | +1.85s |
| t=45.5s | ~26.53s | +2.07s |
| t=53-54s peak | ~28.80s | +4.34s |
| t=61.5s, after catch-up silence | ~26.17s | +1.72s |

Normalized backlog growth:

```text
Before:
t≈1s    0.0s   █
t≈15s  +4.5s   █████
t≈30s  +6.2s   ██████
t≈45s +10.0s   ██████████
t≈58s +13.0s   █████████████
t≈75s  +6.6s   ███████

After:
t≈1s    0.0s   █
t≈16s  +0.6s   █
t≈30s  +1.9s   ██
t≈46s  +2.1s   ██
t≈54s  +4.3s   ████
t≈62s  +1.7s   ██
```

Short utterances also emitted partial/final events promptly in the post-change run, including sub-second reactions such as `うん?`, `えっ!?`, `うわ!`, and `でも`.

### DIFF
--- a/examples/cli/crispasr_run.cpp
+++ b/examples/cli/crispasr_run.cpp
@@ -1424,15 +1424,20 @@ int crispasr_run_backend(const whisper_params& params_in) {
                 // machine *after* this loop runs, so it reflects the
                 // upper bound of all utterances finalized before this step.
                 const int64_t window_start_sample_now = cumulative_samples - (int64_t)pcm_window.size();
+                constexpr int kStraddleMinSamples = 32000; // 2 s @ 16 kHz; backend-safe tail decode floor.
                 for (const auto& sl : slices) {
-                    auto slice_segs =
-                        backend->transcribe(pcm_window.data() + sl.start, sl.end - sl.start, sl.t0_cs, params);
+                    const int64_t s_start_abs = window_start_sample_now + (int64_t)sl.start;
+                    const int64_t s_end_abs = window_start_sample_now + (int64_t)sl.end;
+                    if (params.stream_json && s_end_abs <= finalized_until_sample)
+                        continue;
+
+                    std::vector<crispasr_segment> slice_segs;
                     if (params.stream_json) {
                         // Apply punc/strip on a copy so the per-slice text
                         // is in its final form before we hand it to the
-                        // utterance state machine. The aggregate gets the
-                        // same treatment after the loop, so plain-text
-                        // mode behavior is unchanged.
+                        // utterance state machine. JSON mode can filter or
+                        // trim finalized rolling-window slices before decode;
+                        // plain-text mode below still decodes full slices.
                         std::vector<crispasr_segment> sl_for_text;
 
                         // Round 3 (CKwasd #1 corner): if the VAD slice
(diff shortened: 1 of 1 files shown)


CrispStrobe (approved):
Logic is correct after tracing all three slice paths (finalized, straddling, new) and the non-JSON fallback.

Key improvements over the old code:
- Finalized slices: zero encode passes instead of one per step.
- Straddling slices: one tail encode instead of two (full + tail re-decode). `slice_segs` now carries the tail result, so the aggregate `segs` no longer accumulates already-finalized speech.
- Short-tail straddle (< 2 s): both `sl_for_text` and `slice_segs` stay empty — better than the old code where the full stale decode still went into `segs`.
- Non-JSON mode unchanged via the new else branch.

Minor nit: `s_end_abs` is computed unconditionally but only read when `params.stream_json` is true. No functional impact.

The A/B numbers (+13 s to +4.3 s backlog growth, 67% reduction) are persuasive. Good change.

PR #92, opened 2026-05-15
perf(stream): skip aggregate segs in JSON VAD path

CKwasd:
## Summary

Clean up unused aggregate segment work in the live `--stream-json --vad` path.

In this mode, output is driven by `step_slice_text` and the JSON utterance state machine. The legacy aggregate `segs` vector is not consumed by the JSON path, so this avoids building and post-processing it there.

No intended transcript behavior change.

## What Changed

- Keep per-slice JSON text generation unchanged.
- Avoid appending JSON+VAD slice results into aggregate `segs`.
- Skip aggregate post-processing for JSON+VAD mode.
- Preserve aggregate `segs` behavior for non-JSON streaming and JSON streaming without VAD.
- Use an explicit `decoded_segments_this_step` flag for stream monitor silence detection, since `segs` is intentionally empty in JSON+VAD mode.

## Why

The old flow did extra work that JSON mode did not consume:

```text
VAD slice
  -> backend decode
  -> build per-slice text for JSON state machine
  -> also append same segments into aggregate segs
  -> postprocess aggregate segs
  -> JSON path continues before legacy text output uses segs
```

The new flow keeps only the data needed by JSON mode:

```text
VAD slice
  -> backend decode
  -> build per-slice text for JSON state machine
  -> JSON state machine consumes step_slice_text
```

## Validation

- Built locally.
- `git diff --check`: pass.
- File-mode smoke test with Cohere JA + FireRed VAD + FireRed punctuation on a
  150.9s MP3 completed successfully.
- Live `--stream-json --vad --stream-step 500` smoke test passed with the same
  150.9s MP3 piped as raw `s16le` PCM through `cmd.exe`:
  - emitted `partial` and `final` JSON events;
  - `utterance_id` advanced through the full audio;
  - final silence timestamp reached `150.869`, matching the input duration.

### DIFF
--- a/examples/cli/crispasr_run.cpp
+++ b/examples/cli/crispasr_run.cpp
@@ -1461,15 +1461,14 @@ int crispasr_run_backend(const whisper_params& params_in) {
             }
 
             std::vector<crispasr_segment> segs;
-            // Per-slice text, kept alongside the aggregate `segs` for the
-            // round-2 JSON state machine. Each entry is `(slice, text)`
+            // Per-slice text for the JSON state machine.
+            // Each entry is `(slice, text)`
             // where the text has had `apply_punc_model` + (optionally)
-            // strip-punctuation applied, matching what the aggregate
-            // `segs` will look like after the post-loop processing —
-            // the aggregate is the source of truth for the legacy plain
-            // text path; this side-channel exists so the JSON path can
-            // attribute each slice's text to the right utterance.
+            // strip-punctuation applied. In VAD+JSON mode this is the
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Thank you @CKwasd! Sorry this sat open as long as it did — the diff was integrated into main on a follow-up commit (6318884c) at the time, but we should have pressed the merge button to give you proper credit. Done now. Your test-aware refactor of the JSON+VAD aggregate path is still load-bearing in main, and we layered a small `StreamJsonVadRoutingTests` follow-up on top of it in 42494f23 to lock in the invariants you laid out.

PR #96, opened 2026-05-17
fix(server): Output timestamps + faster responses (works on my computer)

SegFaultZero:
- use shared slicing and forced alignment for timestamped responses
- Use VAD
- Don't call aligner if response_format doesn't require the data
- Faster responses for Qwen3-ASR and Qwen3-Forced-Aligner on my computer

### DIFF
--- a/examples/cli/crispasr_server.cpp
+++ b/examples/cli/crispasr_server.cpp
@@ -24,6 +24,7 @@
 #include "crispasr_output.h"
 #include "crispasr_model_mgr_cli.h"
 #include "crispasr_vad_cli.h"
+#include "crispasr_aligner_cli.h"
 #include "whisper_params.h"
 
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Cherry-picked as ab31e078 into main. Reviewed and verified:
- `crispasr_compute_audio_slices` (VAD+energy fallback) is already in the crispasr link closure of the server binary — no CMakeLists change needed.
- `crispasr_aligner_cli.h` is already compiled into the same target.
- `need_timestamps` correctly skips the aligner for `json` and `text` response formats.
- Live test with qwen3-asr Q4_K on JFK audio passes.

Follow-up: added 8 new unit tests (`test_slice_timestamps`, test #146–153) that pin the t0_cs/t1_cs math `crispasr_fixed_chunk_slices` produces — these are the timestamp offsets passed to each backend call after this change.

PR #97, opened 2026-05-17
fix(qwen3): preserve token ids in beam-search segments

SegFaultZero:


CrispStrobe:
Cherry-picked as 1f588ab8 into main. Verified: qwen3 beam-search now sets `tk.id = id`, matching the greedy path at line 503. Built clean; 164 unit tests pass including two new beam token-ID invariant tests (test #144–145) that pin this contract going forward.

Issue #114, opened 2026-05-19
[bug] Parakeet JA transcription regression post-v0.6.6 - missing segments, kanji degradation and performance


exn251:
Transcription quality for Parakeet JA is noticeably worse on the latest commit compared to the official v0.6.6 release. The regression is also present in v0.6.7

Kanji degradation:
The new build outputs hiragana where the the 0.6.6 release correctly produces kanji. 
Post-v0.6.6: なまえもおしえてくれるかいぼくのなまえは
v0.6.6: 名前も教えてくれるかいぼくの名前は。

Missing segments (segments present in v0.6.6 but dropped entirely in the latest commit build):

00:00:09 — やよっおかゆと申します。
03:08 — room description segment (おかゆ、どう、新しい部屋は...)
04:43 — 十花のジムが映ってるわ、パパが出るかもよ...
05:53 — 小田牧ポケモン研究所。
06:37 — おかゆです小玉博士のいえ。
07:24 — アハハハ、やべえ、やばい...
08:49 — はーいはるかちゃん...
09:27 — ジグザグマになってるサファイア... segment

Might be related that the performance is different now as well:
Newest commit build: 53.23s → 11.3x realtime
Build v0.6.6 : 40.87s → 14.7x realtime (faster and better quality)

GitHub release v0.6.6:
```log
D:\CrispASR\Bench>crispasr.exe -f 10-min-parakeet-test.wav -m parakeet-tdt-0.6b-ja.gguf -l ja -osrt -vm firered
crispasr 0.6.6 (git 5f1bb858, Release) [backends: cpu,vulkan]
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = AMD Radeon RX 5600 XT (AMD proprietary driver) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 32768 | int dot: 0 | matrix cores: none
crispasr: auto-detected backend 'parakeet' from 'parakeet-tdt-0.6b-ja.gguf'
parakeet: vocab=3072  d_model=1024  n_layers=24  n_heads=8  ff=4096  pred=640  joint=640
parakeet: BN folded into conv_dw weights for 24 layers
crispasr: audio: 9599896 samples (600.0 s) @ 16000 Hz, 4 threads
crispasr[vad]: using cached C:\Users\user/.cache/crispasr/firered-vad.gguf
crispasr: processing 56 slice(s)
[...]

Issue #89, opened 2026-05-14
Parakeet Japanese, chunks and missing words


lenhone:
`parakeet-tdt-0.6b-ja.gguf` is not transcribing all or most spoken words. The amount of words transcribed varies greatly with chunk-seconds. I tested 10s, 30s (default) and 180s. There also appears to be an issue where the start of chunk sometimes isn't transcribed. 

Code used
```
   ffmpeg -ss 00:00:00 -i $videopath -ar 16000 -ac 1 -t 300 -c:a pcm_s16le -f wav - | \
   ~/CrispASR/build/bin/crispasr \
    --gpu-backend vulkan \
    --device 1 \
    -m $model \
    -l ja \
    -osrt \
    --output-json-full \
    --split-on-punct \
    --flush-after 1 \
    -ml 1 \
    --chunk-seconds 10 \
    -f - \
```

The below graph hopefully shows the issue clearly

<img width="613" height="409" alt="Image" src="https://github.com/user-attachments/assets/4e37c819-9a4b-4b55-9261-5b15aa5787f5" />

CrispStrobe:
Thanks. Addressed in commit 617cd02 (just pushed).

**Root cause — two separate issues:**

1. **Encoder context**: FastConformer uses full bidirectional self-attention within each chunk. Shorter chunks mean boundary frames see less total context → quality degrades at chunk edges. This explains the chunk-size dependency in your graph.

2. **Filtering bug** (words dropped at slice boundaries): Each slice was extended ±2 s before transcribing (to give the encoder context), then the result was trimmed back to the original window. The trim used a segment-level filter `seg.t0 < sl.t0_cs`, but parakeet sets `seg.t0 = first_word.t0`, which always lands in the left-context region when context is added. So every slice after the first was silently dropped entirely.

**Changes:**

- Each slice is still transcribed with ±2 s of audio context so boundary frames have better encoder input.
- Trimming is now word-level: each word is assigned to the slice where it **starts** (`w.t0 < sl.t0_cs || w.t0 >= sl.t1_cs`). Segment text is rebuilt from the surviving words. This is consistent for parakeet because TDT decoding is frame-synchronous — the same spoken frame always maps to the same centisecond timestamp regardless of which context window it appears in.

[...]

PR #116, opened 2026-05-21
Fix VAD chunk-context handling and zero-duration subtitle timestamps

CKwasd:
## Summary

This PR fixes two subtitle regressions in the non-Whisper ASR dispatcher:

- VAD-derived slices could still enter the chunk-context/LCS trimming path, causing valid Japanese utterances to be merged or dropped.
- Granite beam-search results did not initialize segment timestamps, which could produce zero-duration SRT entries after downstream alignment/output processing.

The fix restores stable Japanese SRT output for Cohere, Granite Speech 4.1, Parakeet, FunASR, and SenseVoice test runs.

### VAD slices were misclassified as chunked audio

The overlap-save chunk-context path is useful for fixed `--chunk-seconds` slicing because those slices are arbitrary cuts through continuous audio. It is not appropriate for VAD-derived slices, which already represent speech boundaries.

The existing gate only checked `effective_chunk_seconds > 0`. That was insufficient for backends such as Cohere, where the CLI fallback chunk size remains non-zero even when VAD produced the actual slices. As a result, VAD segments could still be extended with neighboring audio and then trimmed/deduplicated by LCS, collapsing valid subtitle segments.

### Granite beam-search segments had no timestamps

Granite's greedy path initialized segment start/end times from the input slice, but the beam-search path did not. When Granite ran through beam search, downstream subtitle and aligner logic could receive zero-initialized segment timing.

### Forced alignment accepted unusable spans

The dispatcher accepted any non-empty CTC alignment result. Some aligner outputs can contain words without a positive time span. Those results should not replace valid segment-level timing.

## Changes

- Add an explicit `vad_slicing` input to `should_use_chunk_context()`.
- Disable chunk-context/LCS overlap handling whenever slices came from VAD.
- Pass the dispatcher `wants_vad` state into the chunk-context gate.
- Initialize Granite beam-search segment `t0/t1` from the input audio slice.
- Preserve Granite beam-search token confidence metadata.
- Only apply forced-alignment timestamps when the returned words have a positive time span.
- Update chunk-context gate coverage for the case where VAD slicing is active but `effective_chunk_seconds` is still non-zero.

## Compatibility with #89 and #114

This does not revert the #89 chunk-overlap work. Overlap-save context and LCS stitching still run for real fixed-chunk processing, including explicit `--chunk-seconds` and the long-audio fallback path used when VAD is off.

The change only prevents VAD-derived slices from entering that path. That matches the #114 invariant: VAD slices are already speech-bounded segments, not arbitrary chunk boundaries, so they should not pull neighboring utterances into the decoder context or be LCS-deduplicated as if they were overlapping chunks.

In short:

- `--vad`: no chunk-context overlap, no chunk-boundary LCS dedup.
- `--chunk-seconds` / long-audio fallback without VAD: chunk-context overlap remains enabled.
- Valid CTC forced-alignment spans are still used; only non-positive alignment spans are ignored.

## Verification

Built successfully with:

```powershell
Y:\AI\ASR\CrispASR\build-vulkan.bat
```

Regression audio:

```text
Y:\AI\ASR\1-RJXXXXXX.mp3
```

### Before

On current `main`, the same Cohere Japanese + VAD regression case collapsed valid VAD-derived subtitle segments into a much smaller merged output:

```text
cohere-transcribe-ja-v0.1 + VAD
segments: 8
```

Granite Speech 4.1 with VAD and the Qwen3 forced aligner also produced repeated zero-duration subtitle timestamps:

```text
granite-speech-4.1-2b + VAD + Qwen3 forced aligner
zero-duration subtitle lines: 25
```

### After

With this patch and the rebuilt Vulkan CLI:

| Model | Exit | Segments | Zero-duration subtitle lines |
|---|---:|---:|---:|
| cohere-transcribe-ja-v0.1 + VAD | 0 | 33 | 0 |
| granite-speech-4.1-2b + VAD | 0 | 33 | 0 |
| granite-speech-4.1-2b + VAD + Qwen3 forced aligner | 0 | 36 | 0 |

### Broader Retest

Also retested the Japanese model set; all completed successfully with zero zero-duration subtitle lines:

| Model | Exit | Segments | Zero-duration subtitle lines |
|---|---:|---:|---:|
| cohere-transcribe-original + VAD | 0 | 33 | 0 |
| parakeet-tdt-0.6b-ja + VAD | 0 | 36 | 0 |
| funasr-nano + VAD | 0 | 91 | 0 |
| funasr-mlt-nano + VAD | 0 | 80 | 0 |
| sensevoice-small + VAD | 0 | 66 | 0 |

The Cohere Japanese VAD regression no longer collapses the file to a small number of merged segments, and Granite no longer emits zero-duration SRT timestamps in the tested paths.

### DIFF
--- a/examples/cli/crispasr_backend_granite.cpp
+++ b/examples/cli/crispasr_backend_granite.cpp
@@ -335,11 +335,22 @@ public:
                 free(text);
 
             crispasr_segment seg;
+            seg.t0 = t_offset_cs;
+            seg.t1 = t_offset_cs + (int64_t)((double)n_samples / 16000.0 * 100.0);
             // Strip leading space
             size_t start = 0;
             while (start < transcript.size() && transcript[start] == ' ')
                 start++;
             seg.text = transcript.substr(start);
+            seg.tokens.reserve(gen_ids.size());
+            for (size_t i = 0; i < gen_ids.size(); i++) {
+                if (gen_ids[i] == eos_tok)
+                    break;
+                crispasr_token ct;
+                ct.id = gen_ids[i];
+                ct.confidence = (i < probs.size()) ? probs[i] : -1.0f;
+                seg.tokens.push_back(std::move(ct));
+            }
             if (!params.punctuation) {
                 crispasr_strip_ascii_punctuation(seg.text);
                 crispasr_lowercase_ascii(seg.text);
--- a/examples/cli/crispasr_chunk_context_gate.h
+++ b/examples/cli/crispasr_chunk_context_gate.h
@@ -15,8 +15,9 @@
 // on parakeet-tdt-0.6b-ja was kanji collapsing to bare hiragana plus
 // entire short slices being dropped.
 //
-// The fix is to gate overlap-save on `effective_chunk_seconds > 0`:
-// it's a chunking-only mitigation, and VAD slicing is not chunking.
+// The fix is to gate overlap-save on the actual slice source:
+// it's a chunking-only mitigation, and VAD slicing is not chunking even
+// when the CLI's fallback chunk_seconds value is non-zero.
 
 #pragma once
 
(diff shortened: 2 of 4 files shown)


CrispStrobe:
Thanks. Cherry-picked into main as 1037bcb (merged alongside the #89 30s chunk fix in bdc8175). Both fixes verified together — 31 test cases, 355 assertions pass. Live-tested on the reporter's exact audio (o_9dWkRPYC0): parakeet-JA now produces full 60s/300s coverage where it previously emitted 0 chars at 60s chunks.

PR #134, opened 2026-05-31
fix(test): correct qwen3 backend name + stub-model guard in session-beam suite

CrispStrobe:
Found by actually building + running `test-session-beam` on real models (M1/Metal) — the parallel-session commits (`ef3c37e4` + `3a04b672`) were syntax-checked only.

**Bug 1 — qwen3 cases never opened a session.** They called `open_explicit(model, "qwen3-asr", …)`, but the dispatch string is `"qwen3"` (GGUF arch `qwen3-asr` → backend `"qwen3"`; `transcribe_single` matches `s->backend == "qwen3"`). So `open_explicit` returned `nullptr` and the cases always failed at `REQUIRE(s != nullptr)`, never exercising the `run_with_probs` replay path they target. Fixed at all 3 call sites.

**Bug 2 — no-regression checks passed vacuously on a stub model.** They asserted only `text1 == text2(==text3)`; on a tensorless fixture (`for-tests-ggml-*.bin` loads with *no tensors → empty model for testing*), empty == empty is green. Added `REQUIRE_FALSE(text1.empty())` to all three so a stub fails loudly.

**Verified on real models** (`ggml-tiny.en.bin`, `qwen3-asr-0.6b-q4_k`, `glm-asr-nano-q4_k`): unit 2/2, live 7/7 (49 assertions), all green — whisper native, qwen3-asr replay, and glm-asr setter beam paths confirmed.

### DIFF
--- a/tests/test-session-beam.cpp
+++ b/tests/test-session-beam.cpp
@@ -197,6 +197,7 @@ TEST_CASE("beam: whisper greedy no-regression (beam_size=1 == default)", "[beam]
     INFO("default:    " << text1);
     INFO("beam_size=1:" << text2);
     INFO("beam_size=0:" << text3);
+    REQUIRE_FALSE(text1.empty());  // stub-model guard: empty==empty must not pass vacuously
     REQUIRE(text1 == text2);
     REQUIRE(text1 == text3);
 }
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Added canary + cohere AED live cases (`c27b1dc2`) — covers the Track B branched-KV beam (`52cfec83`): greedy no-regression + beam=2 non-empty for both, gated on `CRISPASR_MODEL_CANARY` / `_COHERE`.

⚠️ **These two backends are compiled + discovered but NOT locally live-run**: loading cohere (1.5 GB) + canary (705 MB) with per-beam self-attention KV snapshots OOMs my dev box. `test-session-beam --list-tests '[.live]'` confirms all 11 cases are present; whisper/qwen3-asr/glm-asr were validated green (7/7, 49 assertions). **Please run canary/cohere on a higher-RAM host or in CI** to confirm the AED beam produces output and stays greedy-bit-identical at beam_size=1.

PR #149, opened 2026-06-05
fix(cli): skip companion download prompt when --codec-model is set (#148)

### LINKED ISSUE #148: mimo-asr: companion download prompt ignores --codec-model / local model dir
## Description

When running mimo-asr with a manually-specified --model and --codec-model pointing to local files, the CLI still prompts to download the companion tokenizer before backend init has a chance to use the provided --codec-model path.

## Reproduction

`powershell
.\crispasr.exe --model D:\models\MiMo-ASR-Q4_K.gguf ^
  --codec-model D:\models\MiMo-Audio-Tokenizer-Q4_K.gguf ^
  --language zh -f audio.flac
`

Observed:
1. CLI auto-detects backend as 'mimo-asr' from the GGUF
2. **Companion pre-download logic** (in xamples/cli/crispasr_run.cpp ~L1172-L1183) kicks in **before** backend init():
   - It looks up the registry entry for mimo-asr, finds companion_filename = "mimo-tokenizer-q4_k.gguf"
   - Calls crispasr_resolve_model_cli("mimo-tokenizer-q4_k.gguf", ...) which checks only current dir + cache dir
   - File not found → prompts "Download now? [Y/n]" with wrong size label (~4.2 GB instead of ~395 MB)
3. Even if user answers 'n', backend init() later properly discovers the tokenizer from --codec-model and works fine

## Root cause

The companion download logic at crispasr_run.cpp:1172-1183 runs before backend init, so it cannot know about:
- The user-supplied --codec-model path (params.tts_codec_model)
- The model-file sibling directory search that discover_audio_tokenizer() in the backend adapter performs

## Expected behavior

The companion pre-download should:
1. First check if --codec-model is already provided (skip companion download if so)
2. If not provided, check the model file's directory before prompting
3. Use the correct size label (~395 MB for tokenizer, not ~4.2 GB)

## Additional notes

The companion size is also wrong in the prompt — it shows the LM model's size (~4.2 GB) instead of the tokenizer's (~395 MB). This is because crispasr_resolve_model_cli prints from the matched entry, and when searching by filename the entry still carries the LM's metadata rather than the companion's.

## Environment
- crispasr 0.6.12 (git b5fbfaa2)
- Windows 11
- CUDA + Vulkan backends

Qxy0happy:
## Problem

When running mimo-asr with \--codec-model\ pointing to a local tokenizer file, the CLI still prompts to download the companion before backend init:

\\\
crispasr: model 'mimo-tokenizer-q4_k.gguf' not found locally.
  Available for download: mimo-tokenizer-q4_k.gguf (~4.2 GB)
  Download now? [Y/n]
\\\

The companion pre-download in \crispasr_run.cpp\ runs before backend init, so it cannot see that the user already provided \--codec-model\. The backend init later uses the explicit path and works fine.

## Fix

Add \params.tts_codec_model.empty()\ to the companion pre-download guard — skip it entirely when the user has already specified the codec path. One line.

Closes #148.

### DIFF
--- a/examples/cli/crispasr_run.cpp
+++ b/examples/cli/crispasr_run.cpp
@@ -1169,7 +1169,7 @@ int crispasr_run_backend(const whisper_params& params_in) {
     // --auto-download set. Doing it here in the dispatcher covers every
     // current and future backend uniformly. Companion lands in the same
     // cache_dir as the LM so the backend's local `discover_*` finds it.
-    if (!backend_name.empty()) {
+    if (!backend_name.empty() && params.tts_codec_model.empty()) {
         CrispasrRegistryEntry entry;
         if (crispasr_registry_lookup(backend_name, entry, params.model_quant) && !entry.companion_filename.empty()) {
             const std::string resolved_companion =
--- a/examples/cli/crispasr_run.cpp
+++ b/examples/cli/crispasr_run.cpp
@@ -1172,11 +1172,22 @@ int crispasr_run_backend(const whisper_params& params_in) {
     if (!backend_name.empty() && params.tts_codec_model.empty()) {
         CrispasrRegistryEntry entry;
         if (crispasr_registry_lookup(backend_name, entry, params.model_quant) && !entry.companion_filename.empty()) {
-            const std::string resolved_companion =
-                crispasr_resolve_model_cli(entry.companion_filename, backend_name, params.no_prints, params.cache_dir,
-                                           params.auto_download, params.model_quant);
-            if (params.verbose) {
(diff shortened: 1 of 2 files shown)


CrispStrobe:
Superseded by f909aa29 on main which takes a more thorough approach: three-tier local check (`--codec-model` / sibling / cache), `companion_approx_size` for all 32 entries, and 43 new unit + integration tests. Thanks for the report, the initial fix, and for flagging the size label issue @Qxy0happy — it helped shape the proper fix!

PR #160, opened 2026-06-10
conv1d transpose: efficient decomposed mul_mat + col2im_1d (~9× codec speedup)

Rafa00127:
# conv1d transpose: efficient decomposed mul_mat + col2im_1d (codec ~1200ms → ~130ms)

I've been trying to fix the conv1d transpose issue for the past few days(I'm not sure if you still remember the issue I posted about, but it's been bothering me a lot since then.), and then I came across a really elegant solution in [qwentts.cpp](https://github.com/ServeurpersoCom/qwentts.cpp/blob/master/src/causal-trans-conv.h) — it uses **w_perm → mul_mat → col2im_1d → GATHER** to perform a really efficient conv1d transpose. So I ported that approach here and applied it to the Qwen3-TTS codec decoder.

Test results on the same long sentence (10s), AMD 7900 XTX:

**Before — All on GPU (with conv1d transpose falls back on cpu):**
```
codes=2353 ms  codec=396 ms  total=2748 ms  audio=10.4 s  rtf=0.264
codes=2377 ms  codec=392 ms  total=2769 ms  audio=10.4 s  rtf=0.266
codes=2346 ms  codec=416 ms  total=2762 ms  audio=10.4 s  rtf=0.266
codes=2341 ms  codec=413 ms  total=2754 ms  audio=10.4 s  rtf=0.265
codes=2352 ms  codec=388 ms  total=2740 ms  audio=10.4 s  rtf=0.263
```

**Before — All on GPU (Yes, the GPU performs a really slow conv1d transpose — it's even slower than the CPU and takes a lot of time.):**
```
codes=2361 ms  codec=1198 ms  total=3558 ms  audio=10.4 s  rtf=0.342
codes=2333 ms  codec=1197 ms  total=3531 ms  audio=10.4 s  rtf=0.340
codes=2324 ms  codec=1198 ms  total=3522 ms  audio=10.4 s  rtf=0.339
codes=2361 ms  codec=1200 ms  total=3561 ms  audio=10.4 s  rtf=0.342
codes=2316 ms  codec=1198 ms  total=3514 ms  audio=10.4 s  rtf=0.338
```

**After — Decomposed mul_mat + col2im_1d, all weights and computation on GPU:**
```
codes=2380 ms  codec=131 ms  total=2511 ms  audio=10.4 s  rtf=0.241
codes=2357 ms  codec=129 ms  total=2486 ms  audio=10.4 s  rtf=0.239
codes=2442 ms  codec=131 ms  total=2573 ms  audio=10.4 s  rtf=0.247
codes=2393 ms  codec=128 ms  total=2521 ms  audio=10.4 s  rtf=0.242
codes=2343 ms  codec=126 ms  total=2470 ms  audio=10.4 s  rtf=0.237

```

The codec decoder drops from ~1200 ms down to ~130 ms. The talker forward is now the main bottleneck, which is exactly where we want it.

This decomposed approach should benefit any TTS model that relies heavily on ConvTranspose1d upsampling layers — I've only tested on Qwen3-TTS so far since its codec uses them most aggressively, but in general it doesn't hurt to have this path available.

### Metal

`ggml_col2im_1d` only has a CUDA kernel right now. On Metal the `w_perm`-based path will abort with "unsupported op". Someone with Apple hardware needs to port the kernel from `ggml-cuda/col2im-1d.cu` into `ggml-metal.metal` and wire up the dispatch. CPU is unaffected (it doesn't populate `w_perm` and keeps using the old `convt1d_crop` path).

### DIFF
--- a/ggml/include/ggml.h
+++ b/ggml/include/ggml.h
@@ -527,6 +527,7 @@ extern "C" {
         GGML_OP_ROPE_BACK,
         GGML_OP_CLAMP,
         GGML_OP_CONV_TRANSPOSE_1D,
+        GGML_OP_COL2IM_1D,
         GGML_OP_IM2COL,
         GGML_OP_IM2COL_BACK,
         GGML_OP_IM2COL_3D,
@@ -2065,6 +2066,17 @@ extern "C" {
             int                   s0,  // stride
             int                   p0,  // padding
             int                   d0); // dilation
+
+    // col2im_1d: gather columns → 1D signal (ConvTranspose1d decomp).
+    //   col: [K*OC, T_in]  F32/F16/BF16 — GEMM output (mul_mat of pre-permuted weight)
+    //   Output: [T_out, OC] F32  where T_out = (T_in-1)*s0 + K - p0
+    //   op_params: [s0=stride, OC, p0=left_crop]
+    GGML_API struct ggml_tensor * ggml_col2im_1d(
+            struct ggml_context * ctx,
+            struct ggml_tensor  * col, // [K*OC, T_in]
+            int                   s0,  // stride
+            int                   oc,  // output channels (= K*OC / K)
+            int                   p0); // left offset / pre-crop
 
     // CrispASR patch (PR #07-metal-aa-snake-beta): BigVGAN v2 anti-aliased
     // SnakeBeta — fused upsample 2× + sin²(α·x)/β + downsample 2×.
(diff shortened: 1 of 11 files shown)


CrispStrobe:
Merged as 5f600f25 — squashed into a single clean commit. Stripped the ops.cpp mass-reformatting (~4000 lines of whitespace changes), removed debug fprintf spam, fixed dead-code bug in trace callback. Core algorithm is excellent — thank you @Rafa00127!

PR #166, opened 2026-06-13
feat(server): honor --punc-model in server mode

praxeo:
## Problem

`--punc-model` is applied in CLI one-shot mode but **silently ignored under `--server`**. A server started with e.g. `--punc-model fullstop -m parakeet-rnnt-1.1b-q4_k.gguf` still returns lowercase, unpunctuated text from `/v1/audio/transcriptions` and `/inference`, because the FireRedPunc/PCS post-processing step lives in the CLI layer (`crispasr_run.cpp`) and `crispasr_server.cpp` never loads or applies it.

This matters for non-PnC backends (Parakeet RNNT/CTC, etc.) that emit no punctuation natively — there's currently no server-side way to punctuate their output.

## Repro

```bash
crispasr --server --port 8109 -m parakeet-rnnt-1.1b-q4_k.gguf --punc-model fullstop
curl -s -F language=en -F beam_size=1 -F file=@clip.wav \
  http://127.0.0.1:8109/v1/audio/transcriptions
# before: "the patient presents with chest pain and shortness of breath
#          vitals are stable will you order a chest x ray and start oxygen"
# (the same model + --punc-model fullstop via CLI one-shot punctuates correctly)
```

## Fix

Load a resident `fireredpunc_context` once at server start when `--punc-model` is set (reusing the CLI's alias resolution: `auto`/`firered`/`fullstop`/`punctuate-all`, auto-downloaded via `crispasr_cache::ensure_cached_file`), thread it into `do_transcribe`, and apply `fireredpunc_process` per segment right after the existing punctuation-strip block.

- **Opt-in** — no behavior change unless `--punc-model` is passed.
- Gated on the per-request `punctuation` flag (so `punctuation=false` still strips).
- Serialized on a private mutex since the shared punc context isn't re-entrant.
- ~57 lines, one file.

## Verified

Parakeet RNNT 1.1b q4_k, CUDA, warm `/v1/audio/transcriptions`:

- **before:** `the patient presents with chest pain and shortness of breath vitals are stable will you order a chest x ray and start oxygen`
- **after (`--punc-model fullstop`):** `The patient presents with chest pain and shortness of breath. Vitals are stable. Will you order a chest x ray and start oxygen?`

The punctuation pass adds ~100 ms; greedy (`beam_size=1`) end-to-end ~1.1 s.

## Scope / notes

This wires only the **FireRedPunc** path (`auto`/`firered`/`fullstop`/`punctuate-all`). The separate `--punc-model pcs` (XLM-R PCS) path and the CLI's `crispasr_should_auto_enable_punctuation` parity could be added the same way if you'd like — happy to extend in this PR.

### DIFF
--- a/examples/cli/crispasr_server.cpp
+++ b/examples/cli/crispasr_server.cpp
@@ -30,6 +30,8 @@
 #include "crispasr_vad_cli.h"
 #include "crispasr_aligner_cli.h"
 #include "whisper_params.h"
+#include "fireredpunc.h"   // server-mode punctuation restoration (--punc-model)
+#include "crispasr_cache.h" // ensure_cached_file() for resolving the punc model
 
 #include "common-crispasr.h" // read_audio_data
 #include "crispasr_chat.h"   // /v1/chat/completions
@@ -311,7 +313,8 @@ struct transcription_result {
 // Load audio from a multipart file upload, transcribe it, return result.
 // Acquires model_mutex internally.
 static transcription_result do_transcribe(const httplib::MultipartFormData& audio_file, CrispasrBackend* backend,
-                                          std::mutex& model_mutex, whisper_params rp, bool need_timestamps) {
+                                          std::mutex& model_mutex, whisper_params rp, bool need_timestamps,
+                                          fireredpunc_context* punc_ctx = nullptr) {
     transcription_result result;
     result.language = rp.language;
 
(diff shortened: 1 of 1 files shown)


praxeo:
Thanks for picking this up so fast: you are a legend — I see it landed as 36f35f2a, and the §166 parity work (PCS, CTC auto-enable, the language wrappers, streaming) is great. Closing since it's in main.

Issue #172, opened 2026-06-18
[Feature Request] Wyoming protocol support (STT + TTS services) for crispasr-server


metalmon:
## Summary

Add a **Wyoming protocol** endpoint to `crispasr-server` so CrispASR can act as a drop-in **STT and TTS service for Home Assistant** (and any Wyoming client), in addition to the existing OpenAI-compatible HTTP API.

## Motivation

Home Assistant's *Assist* voice pipeline talks to speech services over the [Wyoming protocol](https://github.com/OHF-Voice/wyoming). Today users typically run two separate processes:

- [`wyoming-faster-whisper`](https://github.com/rhasspy/wyoming-faster-whisper) for STT
- [`wyoming-piper`](https://github.com/rhasspy/wyoming-piper) for TTS

CrispASR already provides fast multi-backend ASR (Parakeet/Canary/Whisper/…) **and** 10+ TTS engines in a single binary, with a persistent loaded model and an OpenAI-compatible server. A Wyoming endpoint would let one `crispasr-server` instance **replace both** `wyoming-faster-whisper` and `wyoming-piper` — giving HA users CrispASR's faster transducer ASR and its TTS voices through the protocol they already use, with no Python service stack.

This is purely additive and protocol-standard; it is not specific to any one client.

## What Wyoming is (for scoping)

A peer-to-peer, event-based protocol, typically over TCP. Each message is a line-delimited JSON header — `{"type": ..., "data": {...}, "payload_length": N}\n` — optionally followed by a binary PCM payload. Relevant event groups for an STT/TTS service:

- **ASR:** `transcribe` (+ streamed `audio-start` / `audio-chunk` / `audio-stop`) → `transcript` (optionally streaming `transcript-start` / `transcript-chunk` / `transcript-stop`)
- **TTS:** `synthesize` → `audio-start` / `audio-chunk` / `audio-stop` (PCM out)
- **Discovery:** `describe` → `info` (advertise asr/tts capabilities, models, languages, voices)

## Proposed solution

A `--wyoming-port N` flag (or a `crispasr wyoming` subcommand) that serves the Wyoming wire and maps it onto existing internals:

- **ASR service:** drive the existing streaming session API (`crispasr_session_stream_open` / `crispasr_stream_feed` / `crispasr_stream_get_text`, in `examples/server/ws_stream.*`) from incoming `audio-chunk`s; emit `transcript`. Reuse the existing VAD path (`src/crispasr_vad.cpp`).
[...]

PR #176, opened 2026-06-19
fix(wyoming): real HA data_length framing + language/shutdown/robustness (#172)

metalmon:
Fixes for the new Wyoming server (#172) and follow-ups, **tested end-to-end on Windows + RTX 3060 Ti (CUDA 13.2)** against Home Assistant's official `wyoming` 1.9.0 client. Happy to split this up if you prefer.

## The critical one (C1): transcribe path was dead

HA's official `wyoming` lib (`write_event`) sends a **non-empty `data` object as a separate length-prefixed JSON blob** after the header line (advertised via `data_length`), **not inline**:

```
{"type":"audio-chunk","data_length":M,"payload_length":N,"version":...}\n
<M bytes data JSON>
<N bytes PCM payload>
```

`wyoming_read_header` only read `payload_length` and ignored `data_length`, so for any event carrying `data` (transcribe / audio-start / audio-chunk / synthesize) the server read the data-JSON region as if it were the payload, desynced the stream, failed to parse the next header, and closed the connection. `describe` has empty `data`, so it worked — which is exactly why this slipped through.

**Why the existing test didn't catch it:** `tests/test-server-wyoming.py`'s `wyoming_send` sent `data` *inline* with `payload_length`, mirroring the server's (incorrect) expectation. I changed the test to frame `data` via `data_length` like the real lib — it now fails on the old code and passes on the fix.

## What's in here

| Fix | What |
|---|---|
| **C1** | `wyoming_read_header` reads `data_length`, reads + merges the separate data blob before the payload. Inline `data` still accepted. |
| **C4** | `describe->info` advertises the configured `--language` (asr service/model + voice) instead of hardcoded `"en"`. |
| **C2** | `select()` with a 1s timeout in the accept loop so the listener exits on shutdown (`closesocket()` doesn't unblock a blocking `accept()` on Windows -> `wyoming_stop()` could hang on `join()`). |
| **H3** | Track open client sockets; `wyoming_stop()` `shutdown()`s them to unblock `recv()` and bounded-waits for in-flight detached handlers — avoids backend use-after-free at teardown. |
| **H1** | Accept `width=4` (float32) `audio-chunk` payloads; warn-once + drop other widths instead of silently ignoring. |
| **M1** | Check `WSAStartup()` return; add `WSACleanup()`. |
| test | Faithful `data_length` framing + find Windows `build/bin/Release/crispasr.exe`. |

## Verification (Windows, CUDA, whisper ggml-tiny)

Against the official `wyoming` 1.9.0 client:
- `describe -> info` parsed; `transcribe + audio-start + 172 audio-chunk + audio-stop -> transcript` returns `"And so my fellow Americans ..."` — **no reset**.
- With `--language ru`, `info` advertises `languages: ["ru"]`.

Corrected `tests/test-server-wyoming.py` (now uses real framing): **5/5 pass** (describe/info, info structure, ASR round-trip, mid-stream-disconnect resilience, unknown-event resilience).

## Notes / not changed

- **H4** (model_mutex held for the whole `synthesize`): left as-is — it matches the HTTP server's serialization model; flagging rather than changing locking.
- **C3** (speaker `done_event` between plays): **verified NOT a bug** — miniaudio's `ma_event` is auto-reset on both win32 and posix, so repeated `play()`/`wait()` is correct.
- I left the outgoing server messages sending `data` inline (the official client reads inline `data` fine when `data_length` is absent); only the read path needed the `data_length` handling.

### DIFF
--- a/examples/cli/wyoming.cpp
+++ b/examples/cli/wyoming.cpp
@@ -30,9 +30,11 @@
 #include <algorithm>
 #include <atomic>
 #include <cerrno>
+#include <chrono>
 #include <cmath>
 #include <cstdio>
 #include <cstring>
+#include <mutex>
 #include <string>
 #include <thread>
 #include <vector>
@@ -45,15 +47,18 @@ typedef SOCKET socket_t;
 #define CLOSE_SOCKET closesocket
 #define INVALID_SOCKET_VAL INVALID_SOCKET
 #define SOCKET_ERRNO WSAGetLastError()
+#define SHUTDOWN_BOTH SD_BOTH
 #else
 #include <arpa/inet.h>
 #include <netinet/in.h>
+#include <sys/select.h>
 #include <sys/socket.h>
 #include <unistd.h>
 typedef int socket_t;
 #define INVALID_SOCKET_VAL (-1)
 #define CLOSE_SOCKET close
 #define SOCKET_ERRNO errno
+#define SHUTDOWN_BOTH SHUT_RDWR
 #endif
 
 using json = nlohmann::json;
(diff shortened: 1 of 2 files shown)


CrispStrobe:
Integrated into \`main\` (cherry-picked as 0eae209e). Thanks — C1 was a real, confirmed bug: on \`main\` the corrected test reproduces it (transcribe path desyncs → "Broken pipe"), and it passes after the fix. Validated end-to-end on M1 Metal with ggml-tiny: test-server-wyoming.py 5/5, jfk.wav transcribes.

Two improvements layered on top (7c41d9a4):
- The default whisper language is `auto`, which is not an ISO code HA can route on — `describe→info` now advertises `en` when the configured language is empty or `auto` (the actual transcription language is unchanged, `auto` stays valid there).
- Added Catch2 socketpair unit tests for the data_length framing contract (separate data blob → aligned payload; a payload-only read desyncs; empty-data describe needs no blob) — the framing path previously had only the live test.

Closing as merged.

PR #177, opened 2026-06-19
feat(qwen3-tts): true streaming synthesis (per-chunk PCM) — ~7x faster first audio

metalmon:
True **streaming synthesis** for the qwen3-tts backend — emit PCM as it is generated instead of only returning the whole clip. Built and tested on Windows + RTX 3060 Ti (CUDA). Separate from #176 (that one is the Wyoming fixes); these touch disjoint files.

## Why
qwen3-tts is great but `*_synthesize` returns the full clip, so an interactive/voice-assistant loop waits the whole synth (~rtf 1.2–1.9) before any audio. The talker already generates codes frame-by-frame; this exposes that so the first audio lands almost immediately.

## Result (qwen3-tts-customvoice, voice `sohee`, 6.6 s Russian utterance)
| | streaming | non-stream |
|---|---|---|
| **time-to-first-audio** | **~1.1 s** | ~7.0 s |
| total synth | ~8.0 s (+~15% re-decode) | ~7.0 s |
| audio | 11–42 chunks, identical RMS, continuous (seam jumps far below natural max-delta) |

~7x faster first audio; total only slightly higher from per-window re-decode of the overlap.

## What changed
- **`src/qwen3_tts.{h,cpp}`** — `qwen3_tts_synthesize_streaming(ctx, text, chunk_frames, overlap_frames, cb, user_data, out_n_samples)`. Reuses the AR-loop body verbatim (same prefill / talker + code_pred / KV + `past_hidden`, never reset). Every `chunk_frames` new frames (and at EOS) it decodes a window = last `overlap_frames` emitted frames + new frames via the existing `codec_decode_codes`, discards the overlap pre-roll (`overlap_frames*1920` samples) and fires the callback. Per-window rebuild — **no graph caching** (the `ggml_set_rows` cached-graph CUDA-crash note applies; per-window decode is cheap vs the AR loop).
- **`examples/cli/crispasr_backend.h`** — `crispasr_pcm_stream_callback` + virtual `synthesize_streaming` with a whole-clip fallback, so other backends keep working unchanged.
- **`examples/cli/crispasr_backend_qwen3_tts.cpp`** — override `synthesize_streaming` (chunk_frames=8, overlap_frames=96) + `CAP_STREAMING`; shared voice setup factored into `prepare_synthesis()` (no behavior change).
- **`examples/cli/crispasr_server.cpp`** — `/v1/audio/speech?stream=true` rewired to producer/consumer: a worker synthesizes under `model_mutex` and pushes int16-LE PCM chunks into a CV-guarded queue that the `set_chunked_content_provider` drains as they arrive (intra-sentence). The previous branch buffered the whole synth first, so TTFA stayed at full-synth latency. Non-stream path unchanged.

## Design notes
- **chunk_frames=8** — TTFA is gated by AR generation of N frames, **not** codec decode, so small N wins (chunk=24 regressed TTFA to ~5 s).
- **overlap_frames=96** — codec is causal with sliding-window=72 plus a causal-conv receptive field; overlap=32 left ~6% seam difference vs whole-clip, 96 drops it to ~1.1% (inaudible). Mirrors the existing `transcribe_streaming` / `CAP_STREAMING` / `set_chunked_content_provider` pattern.

## Test
```
crispasr --server --backend qwen3-tts-customvoice -m auto --port 8082
# POST /v1/audio/speech {"input": "...", "voice": "sohee", "stream": true, "response_format": "pcm"}
# stream the response; first PCM chunk arrives ~1.1 s, rest follows continuously.
```

## Known / limitations
- Not bit-exact vs whole-clip (~1.1% at seams) — inaudible; raise `overlap_frames` for stricter fidelity at more re-decode cost.
- No codec graph caching (deliberate, per the CUDA note) — small per-window rebuild overhead.
- Streaming wired for the qwen3-tts backend; other TTS backends fall back to the whole-clip default.

Happy to adjust defaults / split further if you prefer.

### DIFF
--- a/examples/cli/crispasr_backend.h
+++ b/examples/cli/crispasr_backend.h
@@ -154,6 +154,21 @@ public:
         return {};
     }
 
+    // Streaming TTS callback: invoked once per generated PCM chunk as the
+    // backend produces audio, with the last chunk flagged is_final=true.
+    // `pcm` is mono float32 at `tts_sample_rate()`.
+    using crispasr_pcm_stream_callback = std::function<void(const float* pcm, int n_samples, bool is_final)>;
+
+    // Synthesize with streaming output. Default implementation falls back to
+    // the whole-clip synthesize() and emits it as a single final chunk.
+    // Backends with CAP_STREAMING override to emit incrementally.
+    virtual void synthesize_streaming(const std::string& text, const whisper_params& p,
+                                      crispasr_pcm_stream_callback cb) {
+        auto v = synthesize(text, p);
+        if (!v.empty())
+            cb(v.data(), (int)v.size(), true);
+    }
+
     // Sample rate of `synthesize()` output PCM. Defaults to 24 kHz since most
     // TTS backends (kokoro, qwen3-tts, vibevoice, chatterbox, orpheus, indextts)
     // produce 24 kHz. Backends that emit a different rate (e.g. voxcpm2-tts at
--- a/examples/cli/crispasr_backend_qwen3_tts.cpp
+++ b/examples/cli/crispasr_backend_qwen3_tts.cpp
@@ -90,7 +90,7 @@ public:
     const char* name() const override { return "qwen3-tts"; }
 
     uint32_t capabilities() const override {
-        uint32_t caps = CAP_TTS | CAP_AUTO_DOWNLOAD | CAP_TEMPERATURE | CAP_FLASH_ATTN;
+        uint32_t caps = CAP_TTS | CAP_AUTO_DOWNLOAD | CAP_TEMPERATURE | CAP_FLASH_ATTN | CAP_STREAMING;
         if (is_base_)
             caps |= CAP_VOICE_CLONING;
         return caps;
@@ -147,9 +147,13 @@ public:
         return true;
     }
 
-    std::vector<float> synthesize(const std::string& text, const whisper_params& params) override {
+    // Shared voice/customvoice/voicedesign setup for both synthesize() and
+    // synthesize_streaming(). Configures temperature, seed, language, and the
+    // active voice identity (re-loading only when the identity changes via
+    // last_voice_key_). Returns true when the context is ready to generate.
+    bool prepare_synthesis(const std::string& text, const whisper_params& params) {
         if (!ctx_ || text.empty())
-            return {};
+            return false;
         qwen3_tts_set_temperature(ctx_, params.temperature);
         qwen3_tts_set_seed(ctx_, params.seed);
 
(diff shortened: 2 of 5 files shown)


metalmon:
Two clarifications:

**1. The streaming framework is backend-agnostic — qwen3-tts is just the first implementation.**
`synthesize_streaming` (virtual), `CAP_STREAMING`, and the `/v1/audio/speech?stream=true` producer/consumer wiring are all backend-independent. Other autoregressive / frame-wise TTS backends (csm, dia, bark, vibevoice, orpheus, …) can opt in by overriding `synthesize_streaming` the same way qwen3-tts does. Non-AR backends keep the whole-clip default fallback, so nothing regresses. Happy to wire up another backend in this PR if you'd like a second example.

**2. Precise note on graph caching / CUDA (correcting my earlier wording).**
The per-window codec decode rebuilds its graph each window (no caching). I avoided caching **conservatively**, not because it's proven broken: the existing `O15` cached-graph reuse in `code_pred` is already default-OFF on CUDA because of the `ggml_set_rows` `GGML_ASSERT` in `ggml_backend_tensor_set` (issue #56; revert 5e21e4a of b3cd141, Jetson Orin AGX sm_87). That crash is documented for the **code_pred** path, **not** the **codec-decode** graph — so codec-graph caching on CUDA is simply untested here. Per-window rebuild is the safe default and is cheap relative to the talker AR loop, which is the real bottleneck and what gates time-to-first-audio. If you want, I can add opt-in codec-graph caching (env-gated like O15) and test it on CUDA.

CrispStrobe:
Integrated into \`main\` (cherry-picked as 8c1f853f), then hardened on top. Thanks — this is a solid feature. Validated on M1 Metal (qwen3-tts-0.6b base + voice-default):
- **TTFA ~5.8× faster** (≈9.3s → ≈1.6s), output **ASR-identical** to the whole-clip path (max seam diff ~1.6%, both transcribe verbatim).
- One platform caveat documented: on Metal the codec re-decode is **not** cheap relative to the AR loop, so total wall time ~doubles (~1.9×) vs the documented ~15% on CUDA. The win is interactive latency, not throughput — callers wanting minimal total time on Metal should keep stream=false.

Two follow-ups layered on (352adcee, cd2ea5c7):
- **Server robustness**: the "bounded queue" had no cap and the worker was detached — a client disconnecting mid-stream left it synthesizing under model_mutex and growing the deque. Added a real backpressure cap + a cancel flag the chunked provider sets on write failure (worker stops promptly), and fixed a latent dangling `[&]` capture (push_pcm is copied into the detached thread). Verified: mid-stream disconnect leaves the server responsive.
- **Dedup**: synthesize_streaming duplicated ~250 lines of the talker AR loop from synthesize_codes. Extracted into one template helper with a per-frame hook (net −162 lines), gated on byte-identical `--tts` output so the whole-clip path is provably unchanged.
- Registered test-server-tts.sh as a live ctest and added a streaming section (stream=true pcm/wav→200, mp3→rejected, log marker).

Closing as merged.

PR #180, opened 2026-06-21
fix(cmake): try cblas library before blas for linking

BergmannAtmet:
`src/CMakeLists.txt` checks for `cblas.h` and the blas library
 separately.

 But `cblas.h` is a part of cblas, not blas.

 This can result in linking errors, as a reference to `cblas_sgemm`
 is not found, because libblas doesn't have the symbol:

     mel.cpp:(.text+0x19a8): undefined reference to `cblas_sgemm'

 This commit looks for the cblas library first, and links against it
 if found. Otherwise, the old behavior of trying blas is done.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -288,7 +288,13 @@ else()
     # so this find_package(BLAS) call is steered toward MKL automatically.
     # cohere.cpp also does a bare `#include <cblas.h>` under HAVE_BLAS — gate on
     # the header being reachable (see crispasr-core note above re: MSYS2).
-    find_package(BLAS)
+    find_package(CBLAS)
+    if (CBLAS_FOUND)
+      set(BLAS_FOUND ON)
+      set(BLAS_LIBRARIES ${CBLAS_LIBRARIES})
+    else()
+      find_package(BLAS)
+    endif()
     find_path(CBLAS_INCLUDE_DIR cblas.h PATH_SUFFIXES openblas)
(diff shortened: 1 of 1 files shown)


BergmannAtmet:
**Note:** I don't do cmake, and don't know if this is a *proper* fix. You can consider it an issue submitted as a PR, if you wish.

CrispStrobe:
Thanks @BergmannAtmet — good catch. The split-BLAS link failure (`cblas_*` lives in a separate `libcblas` on reference/Netlib BLAS, while `find_package(BLAS)` only resolves the Fortran `libblas`) is a real bug.

Merged your commit to `main` (with attribution preserved), plus a follow-up that makes it actually take effect everywhere: stock CMake ships `FindBLAS` but **no `FindCBLAS`**, and the repo didn't provide one — so on a bare Debian/Netlib box `find_package(CBLAS)` silently finds nothing and the fix degrades to a no-op (it only worked where a `CBLASConfig.cmake` happened to be installed). The follow-up:

- adds `cmake/FindCBLAS.cmake` (`find_library(NAMES cblas)`) so `find_package(CBLAS)` resolves a real `libcblas`;
- DRYs the two duplicated cohere/mel blocks into a `crispasr_prefer_cblas([QUIET])` macro.

Backward-compatible — OpenBLAS/MKL/Accelerate bundle the cblas symbols, so we fall back to `find_package(BLAS)` exactly as before (verified clean configure on Accelerate-only macOS).

Landed in `main` as 4bb78ec8 (yours) + bd217910. Closing as merged.

PR #188, opened 2026-06-24
Provides a Python-based GUI.

JTNE1221:
Requires the following Python libraries: os, json, shlex, subprocess, threading, and tkinter. Settings can be saved to a JSON file and automatically loaded upon the next launch.

Configurations can be saved manually or automatically during runtime. Supports both Chinese (zh) and English (en) languages ​​(the English text is translated from Chinese, though option descriptions retain the original wording).
Generated by Gemini.（0 is a placeholder file.）


CrispStrobe:
thanks for the idea. the idea of CrispASR is to be a library that is easily wrappable. we already have GUIs 
- in python [Susurrus](https://github.com/CrispStrobe/Susurrus)
- in dart/flutter [CrisperWeaver](https://github.com/CrispStrobe/CrisperWeaver)
which are their own repos. I would suggest you do likewiese and make your GUI its own repo. You might let another more capable coding agent than Gemini rework it though.

BTW i am adding Chinese language support to CrisperWeaver so maybe that works for you. That i can do only by AI assistance as I have no working competence of Chinese. So if you would like to take a look at that and maybe correct the probably several mistakes in AI translations that would be quite helpful.

PR #207, opened 2026-06-29
perf(granite): CUDA-graph-capture bucketed decode

sims1253:
## Summary
Brings CUDA-graph capture to granite-speech's LLM decode, taking it from ~0.9× realtime on the stock path to ~9–13× realtime (bit-identical transcripts).

## Problem
The granite LLM decode was launch-bound. `granite_speech_run_llm_kv` rebuilt the 40-layer cgraph + re-allocated the scheduler graph every token, and grew the KV read length `Lk = n_past+1` each step. That growing shape defeated ggml's CUDA-graph capture (warmup never stabilized), so the ~280 tiny GEMV launches per token stayed as host dispatches. Additionally, every token paid a ~400 KB vocab logits D2H + host-side argmax/softmax scan.

## Solution
Four stacked commits applying the megapar/starling/outetts CUDA-graph lessons:

1. **Bucketed single-token decode graph** (`granite_speech.{cpp,h}`) — pins `Lk` to a fixed bucket covering `[prompt .. prompt+max_new]`, routes KV writes through `set_rows` with a runtime positions index so topology + node shapes are constant. Capture engages after warmup; the whole 40-layer step replays as one `cudaGraphLaunch`.
2. **In-graph argmax** — appends a `ggml_argmax` node to the captured graph; per-token D2H shrinks from ~400 KB vocab → 4 bytes (one int32 id). Greedy selection stays bit-identical (same strict-`>` tie-break).
3. **Fused embed lookup (F16)** — folds the per-token embedding `get_rows` into the same captured graph (capture-safe for non-quantized dtypes), halving graphs/token on F16. k-quant keeps the separate-embed path.
4. **Fixes** — free the cached argmax graph on bucket change (prevents arena leak); document why per-step sched reset+alloc can't be skipped under capture.

Also flips `GGML_CUDA_GRAPHS_DEFAULT` to ON (arch-gated to sm_80+ inside ggml-cuda; `GGML_CUDA_DISABLE_GRAPHS=1` escape hatch), and denies capture for GET_ROWS on k-quant sources (its host sync is illegal inside capture — fixed a Q4_K crash).

## Changes
- `src/granite_speech.{cpp,h}` — cached bucketed decode + argmax-fused + `granite_speech_greedy_decode` API + bucket setter
- `src/core/granite_llm.h` — thread `fixed_kv_len` + `kv_indices` through the attention path
- `examples/cli/crispasr_backend_granite.cpp` — enable the bucket + fast greedy path for the common transcript case
- `ggml/CMakeLists.txt` — `GGML_CUDA_GRAPHS_DEFAULT=ON`
- `ggml/src/ggml-cuda/ggml-cuda.cu` — deny capture for k-quant GET_ROWS
- `docs/contributing.md` — §210: document that host-syncing ops (k-quant GET_ROWS) crash CUDA-graph capture, and the sched-reuse shortcut limitation

## Results
RTX 5090, bf16/f16, model load excluded, median of 5, bit-identical transcripts vs stock path on jfk.wav (11s) and multispeaker.wav (31.5s), F16 + Q4_K:

| clip | before | after | speedup |
|---|---|---|---|
| jfk F16 | 11.8s / 0.9× RTFx | 1.85s / 9.2× RTFx | ~10× |
| jfk Q4_K | ~11s / 1.0× | 1.6s / 9.0× RTFx | ~9× |
| multispeaker F16 | ~9× baseline | 8.3×→12.7× RTFx | 8–13× |

Decomposition (multispeaker F16): bucket graph caching = 4.8×, +capture = 8.2×, +argmax/fuse = 12.7×. Capture wins more on Q4_K (~2.7×) since MMQ kernels capture cleanly.

## Testing
- Bit-identical transcripts vs the stock path: jfk.wav + multispeaker.wav, F16 + Q4_K, capture on/off
- Beam search and all other models untouched (beam keeps the legacy per-step path)
- `ctest -R granite` — 11/11 pass (incl. the new `test-granite-maxlen` from main)
- `./tools/format.sh` — clean under clang-format v18 (CI-pinned); rebased onto latest main with no conflicts
- Full `ninja crispasr-cli` builds clean
- Sampling / best-of-N paths unchanged (still use the shared loop for full logits)

## Context
Beam search and the no-bucket/prefill paths keep the original per-call rebuild, so this only affects the greedy single-run transcript case. Remaining headroom in the the encoder/mel stages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,9 @@ ggml/ggml-version.cmake
 .swiftpm
 *.metallib
 
+# local benchmark models + profiling artifacts
+.bench-models/
+
 ggml-metal-embed.metal
 ggml-metal-embed.metal.tmp
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -435,6 +435,24 @@ not just M1.
   broken cross-backend copy; mid-graph CPU splits are fine. (A `ggml_concat`-of-a-
   scaled-view "manual pad" to avoid `GGML_OP_PAD` fails when pad > input length —
   you can't build zeros wider than the source view.)
+- **Host-syncing ops crash a CUDA-graph-captured graph.** `ggml` can capture and
+  replay an LLM decode step as a single `cudaGraphLaunch` (CUDA-graph capture, on
+  by default via `GGML_CUDA_GRAPHS_DEFAULT`, arch-gated to sm_80+; disable with
+  `GGML_CUDA_DISABLE_GRAPHS=1`), but only if *every* op in the graph is
+  capture-safe. `get_rows_cuda_kquant` copies the index tensor D2H and calls
+  `cudaStreamSynchronize` to read it on the host before launching its per-row
+  dequant kernels — a host sync is illegal inside capture, so a graph containing
+  GET_ROWS on a k-quant source crashes. F16/F32/legacy-quant GET_ROWS use the
+  capture-safe `k_get_rows` path and are unaffected. (This bit
+  `granite-speech-4.1-2b-q4_k.gguf`'s quantized token embedding under capture —
+  §210.) **Fix:** deny CUDA-graph capture for graphs with k-quant GET_ROWS
+  (`ggml_cuda.cu`), and keep the k-quant path on the legacy per-call rebuild. The
+  same rule applies to any future op that host-syncs (e.g. reads a shape/extent on
+  the host before dispatching) — it cannot live in a captured graph; route it
+  through the non-captured path instead. **Related:** even on a topology-stable
+  captured graph you must still `ggml_backend_sched_reset` + `sched_alloc_graph`
+  per step — ggml's capture bookkeeping rejects the documented reuse-shortcut
+  (`ggml-backend.h:285`) with `CUDA error: invalid argument` mid-decode.
 
 ## Watermarking tests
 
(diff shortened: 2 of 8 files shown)


CrispStrobe:
Merged into `main` (c5035969). Thanks @sims1253 — clean, well-documented work.

Validated locally on an M1 (Metal), since the new path is gated on `use_gpu` and so activates on Metal too, not just CUDA:
- `ctest -R granite` → 9/9 green (incl. `test-granite-maxlen`).
- jfk.wav and fleurs_60s.wav: transcripts **byte-identical** between the bucketed GPU path and the legacy CPU path (`set_rows`-bucket + in-graph `argmax` confirmed exercised on Metal via the loaded `kernel_set_rows_f16_i32` / `kernel_argmax_f32` pipelines). Metal doesn't do CUDA-graph capture, but the cached-graph + argmax-fuse alone gave ~2.6× here (0.8× → 2.1× RTFx) with no quality change.

Added one follow-up commit on top (c5035969): the argmax-fused greedy loop wrote the new token's KV at row `n_past` via `set_rows` into the fixed `[0,bucket)` view every step but never checked `n_past` against the bucket. When the CLI's `min(prompt+max_new+1, 4096)` clamp engages and a generation doesn't hit EOS, `n_past` could reach/exceed the bucket → out-of-bounds write (also past the 4096-row KV cache). Now the loop stops at the bucket ceiling (== the hard context limit, where the legacy path also runs out of cache) and the per-step helpers are self-safe. Doesn't trigger on normal clips (verified identical output), pure hardening for the long-generation edge.

sims1253:
@CrispStrobe  Just checking, I've been doing a bunch of these in https://github.com/sims1253/starling that I was planning to slowly port to ggml/CrispASR. Just wanted to double-check if you'd appreciate it or rather not :)

CrispStrobe:
@sims1253 Yes indeed, these look great! Only i have no direct sm_120 access so i cannot cross-check easily.

PR #223, opened 2026-07-06
feat(vibevoice-asr): wire up --context hotword/metadata injection

witwicki:
## Summary

- The HF model card (`hf_readmes/vibevoice-asr-GGUF.md`) advertises `--context "ACME Corp, John Smith, Q3 earnings"` for VibeVoice-ASR hotword/metadata injection, but the flag didn't exist anywhere in the `crispasr` CLI (confirmed via `--help` and `--backend vibevoice --help`).
- Wires `--context TEXT` through `whisper_params` → `crispasr_backend_vibevoice` → `vibevoice_transcribe()`, splicing runtime-tokenized context text into the ASR prompt.
- Matches the real upstream reference rather than guessing: pulled `microsoft/VibeVoice`'s `vibevoice/processor/vibevoice_asr_processor.py`, which has an actual `context_info` parameter producing this exact wording:
  ```python
  if context_info and context_info.strip():
      user_suffix = f"This is a {audio_duration:.2f} seconds audio, with extra info: {context_info.strip()}\n\nPlease transcribe it with these keys: " + ", ".join(show_keys)
  ```
  Confirmed this is the right reference version (not just a plausible-looking file) because its `else` branch (no context) reproduces the existing hardcoded prompt token array in `vibevoice.cpp` byte-for-byte — independent verification that wouldn't hold for a divergent processor version.
- Reuses existing infrastructure rather than adding anything new: the same dual-path text tokenizer already used for VibeVoice TTS text (`core_bpe::tokenize_simple` when the GGUF has `tokenizer.ggml.merges`, else the vocab-rank approximation `tokenize_text_bpe_vocab_rank` — the current released GGUF has no merges table, so the approximation path is what actually runs).
- The no-context path is completely untouched — same hardcoded token arrays as before, zero risk to the previously-verified default prompt.

## Test plan

- [x] `cmake --build build` — clean build, no new warnings
- [x] `ctest --test-dir build -L unit --timeout 30` — 767/767 unit tests pass
- [x] `./build/bin/test-issue-79` — passes with updated call sites (nullptr context, no behavior change)
- [x] Live smoke test against `vibevoice-asr-q8_0.gguf` + `samples/jfk.wav`:
  - No `--context`: 146 prompt tokens, correct transcription
  - `--context ""` / `--context "   "`: identical 146 tokens (matches Python's `.strip()` truthiness check)
  - `--context "ACME Corp, John Smith, Q3 earnings"`: 160 prompt tokens (44 tokens for the context-aware suffix), completes without error, valid JSON output

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -620,7 +620,24 @@ Per-word boost suffix: `"Berenz^5.0,NVIDIA^3.0,plain"`.
 |---|---|
 | **CTC-WS trie (Phase A)** — token-level logit boost during CTC/TDT decode | parakeet (CTC + TDT) |
 | **LLM prompt injection (Phase B)** — hotwords appended to the system/instruction prompt | qwen3-asr, voxtral |
+| **Free-form prompt injection** — separate `--context` flag, not `--hotwords` (see below) | vibevoice |
 | Not applicable | voxtral4b (fixed streaming prompt), granite-nle (NAR, no text prompt), funasr (hardcoded prompt), whisper (use `--prompt` instead) |
+
+### VibeVoice-ASR: `--context` (free-form prompt injection)
+
+VibeVoice-ASR uses its own flag, `--context "TEXT"`, rather than
+`--hotwords` — the model's prompt format accepts free-form prose/metadata
+(names, organizations, topic context) instead of a structured hotword
+list with per-word boost weights. The text is spliced directly into the
+model's instruction prompt, matching the `context_info` parameter in
+microsoft/VibeVoice's `vibevoice_asr_processor.py`. Empty or
+whitespace-only `--context` behaves identically to omitting the flag
+(same prompt, byte-for-byte).
+
+```bash
+crispasr --backend vibevoice -m auto -f meeting.wav \
+    --context "ACME Corp, John Smith, Q3 earnings"
+```
 
 ### Example
 
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -412,6 +412,8 @@ static bool whisper_params_parse_arg_backend_vad(int argc, char** argv, int& i, 
             return false;
         }
         params.lcs_min_length = v;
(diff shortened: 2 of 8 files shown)


witwicki:
Pushed a follow-up commit addressing two potential issues:

1. **ABI break**: `vibevoice_transcribe()`/`vibevoice_transcribe_with_probs()` are `extern "C"` and exported from `libcrispasr.so`. Changing their signatures in place risked breaking any caller still linked against the 3-arg form (undefined 4th-arg register read as a `char*`). Reverted both to their original signatures and added new `vibevoice_transcribe_with_context()` / `vibevoice_transcribe_with_probs_and_context()` entry points instead — purely additive, no ABI risk.

2. **Newline loss in the `--context` prompt**: the injected suffix text embedded `\n`/`\n\n`, but the tokenizer helper (`tokenize_simple`/`tokenize_text_bpe_vocab_rank`) treats whitespace runs purely as word separators and never emits a newline token — silently dropping the leading `\n` (token `198` in the no-context path) and the blank line before "Please transcribe". Now tokenizes the two text lines independently and splices in the structural newlines explicitly (preferring a merged double-newline vocab token when the GGUF has one).

Re-verified: 767/767 unit tests pass, `test-issue-79` passes, and a live smoke test against `vibevoice-asr-q8_0.gguf` confirms the no-context/empty-context/whitespace-context paths are still byte-identical (146 prompt tokens) while `--context` now correctly includes the newline-preserving token structure.

witwicki:
Added documentation: a new subsection under \`## Hotwords / contextual biasing\` in \`docs/cli.md\` covering vibevoice-asr's \`--context\` flag, since it's a distinct free-form-text mechanism from the structured \`--hotwords\` list used elsewhere. Also noted this closes a pre-existing gap — \`docs/architecture.md\` already claimed vibevoice supports "hotwords" before this flag was ever wired up in code.

Issue #217, opened 2026-07-04
Could you please provide a separate alignment mode?


sayou112:
Could you please provide a separate alignment mode? The current workflow first transcribes with an ASR model and then performs alignment with an alignment model. Would it be possible to decouple the latter step so that it can take a .txt file or an unaligned .srt subtitle file as input, run it through the alignment model, and then split and output the corresponding aligned .srt file?

sayou112:
Thank you very much. I noticed this:
--align-only | Standalone forced alignment: text + audio → timestamped SRT/JSON (no ASR needed) |

Will there be support for .srt input as well?

Additionally, I came across another project:
https://github.com/Jerry-at-GH/subtimer

It claims to largely solve long‑sentence segmentation and alignment issues. I'm not sure if it could be helpful for the alignment mode.

Part of its algorithms come from:
https://docs.nvidia.com/nemo-framework/user-guide/latest/nemotoolkit/tools/nemo_forced_aligner.html

It also seems to use an ensemble of three VAD models plus separated speech to compute volume.

CrispStrobe:
`--align-only --text-file subs.srt `preserves your cues and re-emits them with corrected timings; 
`--align-granularity word|segment` controls the output units. 
Our aligner already uses the same CTC Viterbi approach as NeMo Forced Aligner.

sayou112:
I tried --align-only and found that VRAM usage maxes out immediately. Additionally, different models give different errors.

Using parakeet-ctc-1.1b-ja-f16.gguf (from https://huggingface.co/cstr/parakeet-ctc-1.1b-ja-GGUF) can produce output, and alignment for one episode takes about 6 minutes.

With parakeet-ctc-0.6b-GGUF (from https://huggingface.co/cstr/parakeet-ctc-0.6b-GGUF), it errors out:

▶ Aligning: 11_[巨人の星][1969][TV]89(BD 1488x1080 x265 FLAC)_(Vocals).wav
[...]

PR #225, opened 2026-07-06
fix(vibevoice-asr): chunk the encoder to avoid CUDA int32 overflow on long audio

witwicki:
## Summary

- VibeVoice-ASR's encoder crashed with a CUDA int32 overflow (`GELU failed` / `CUDA error: invalid argument`) on audio longer than roughly 700s, because ggml's CUDA unary-op kernels (`ggml/src/ggml-cuda/unary.cu:129`, `unary_cuda`) take the element count as a 32-bit `int`, truncated from `ggml_nelements()`'s `int64_t`. The acoustic/semantic ConvNeXt encoder's stage-0 FFN GELU runs on a `[C_ffn=128, T]` tensor with `T = n_samples` (stride 1, not yet downsampled), so `INT32_MAX/128 ≈ 698.9s` at 24kHz overflows the kernel launch. Reported on a 122GB-VRAM GPU at 1800s audio — a different, VRAM-independent failure mode than a separate OOM ceiling found earlier on a 16GB GPU (~180s).
- Same bug class already fixed for a different backend in `955fbffb` (#217, `src/crispasr_aligner.cpp`).
- Fix lives entirely in `run_encoder_stage()` — the one function shared by every encoder caller (`vibevoice_run_acoustic_encoder`/`vibevoice_run_semantic_encoder`/`vibevoice_encode_speech`'s stage API, `vibevoice_transcribe_impl`, and the TTS voice-cloning reference-encode path), so every caller gets the fix for free:
  - Audio longer than 300s (comfortably under the ~698.9s limit) is split into chunks before calling the encoder graph.
  - **Overlap-and-discard for correctness**: the encoder is causal-only (`build_causal_conv1d`/`build_causal_dw_conv1d` both left-pad only), so a hard cut at a chunk boundary would otherwise feed the first several output frames of every chunk after the first zero history instead of real preceding audio. Each such chunk is extended with 10s of real preceding audio, and the frames covering that extension are discarded before stitching. The discard count is measured empirically (encoding the left-context prefix alone and counting its output frames) rather than computed analytically, since the encoder's stride-alignment right-padding makes an exact closed-form frame count fragile to derive across its 7 ConvNeXt stages.
  - Only the encoder is chunked — the LLM decode still runs as a single pass over the whole concatenated feature sequence for the entire recording, preserving VibeVoice's global context / consistent diarization across long audio (unlike the CLI's outer `--chunk-seconds` dispatch, which restarts the whole model per chunk).
- Also fixed a heap buffer overflow found in review while touching this code: the `VIBEVOICE_REF_FEATURES` debug-injection path (opt-in env var, used for stage-diff testing against a Python reference dump) truncated `T_ref = n / d_lm` when the reference file's float count wasn't an exact multiple of `d_lm`, then allocated the smaller truncated size but read the larger untruncated count into it. Now validated up front and read exactly to the allocated size, failing cleanly on a short read instead of corrupting memory.
- `vibevoice_transcribe_impl` now calls the existing exported `vibevoice_encode_speech()` directly instead of duplicating its own copy of the encoder+connector+sum logic.

## Test plan

- [x] `cmake --build build` — clean build, no new warnings
- [x] `ctest --test-dir build -L unit --timeout 30` — 767/767 unit tests pass
- [x] `./build/bin/test-issue-79` — passes
- [x] `jfk.wav` regression: byte-identical output before/after
- [x] With temporarily-lowered constants (40s chunks / 5s context), a 120s clip needing 3 encoder chunks produced the correct total frame count (900, matching the true single-pass value) and a transcript reading coherently across both internal chunk boundaries with consistent speaker labels and no duplication
- [x] Malformed `VIBEVOICE_REF_FEATURES` file (float count not a multiple of `d_lm`) is now safely rejected with a clear error instead of overflowing; valid reference-file injection still works correctly

### DIFF
--- a/src/vibevoice.cpp
+++ b/src/vibevoice.cpp
@@ -642,9 +642,28 @@ static ggml_cgraph* build_tokenizer_encoder_graph(vibevoice_context* ctx, const 
 // Stage helpers (shared by public stage API and vibevoice_transcribe)
 // ===========================================================================
 
-// Run one tokenizer encoder. Returns [vae_dim * T] row-major, frame-first.
-static std::vector<float> run_encoder_stage(vibevoice_context* ctx, const char* prefix, const float* samples,
-                                            int n_samples, int* out_T, int* out_vae_dim) {
+// Cap per-chunk raw audio length fed to the encoder in one graph-build-
+// and-compute call. ggml's CUDA unary-op kernels (e.g. GELU,
+// ggml/src/ggml-cuda/unary.cu) take the element count as a 32-bit `int`.
+// The encoder's stage-0 FFN GELU runs on a [C_ffn=128, T] tensor with
+// stride 1 (T = n_samples, not yet downsampled), so INT32_MAX/128 ≈ 698.9s
+// @ 24kHz overflows the CUDA kernel launch ("invalid argument"). 300s is
+// comfortably under that, with ~2.3x margin for any other similarly-shaped
+// op, and needs only ~12 chunks for a full 60-minute recording.
+static constexpr int VIBEVOICE_ENCODER_MAX_CHUNK_SAMPLES = 300 * 24000;
+
+// Left-context audio prepended to every chunk after the first, in samples.
+// The encoder is causal-only (build_causal_conv1d/build_causal_dw_conv1d
+// both left-pad only), so a hard cut at a chunk boundary would otherwise
+// feed the first several output frames zero history instead of real
+// preceding audio. 10s is a generous bound on the encoder's true receptive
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Merged with improvements in 388bd80e — cached discard_T (halves encoder calls), env-gated chunk/context constants (CRISPASR_VIBEVOICE_ENCODER_CHUNK_SECONDS, CRISPASR_VIBEVOICE_ENCODER_CONTEXT_SECONDS). Thanks for the fix!

PR #259, opened 2026-07-15
Expose CTC + Whisper introspection accessors on the exception-safe session

mculbert:
## Summary

This bundles several enhancements to the **exception-safe session API**
(`crispasr_session_*`). The common thread: the
exception-safe session is the fault-isolated path a downstream integrator must
use (a native segfault / accelerator OOM stays contained), but several useful
signals were only reachable on the abort-prone legacy `whisper_context` /
per-backend paths. These changes surface them on the session and mirror them
across the language bindings.

Each item is independent; can split into separate PRs if you'd prefer to
review/merge them piecemeal.

### 1. CTC vocab accessor (`Session::ctc_vocab`, comprehensive)
Expose the CTC output vocabulary on the session so the logit grid above is
interpretable (token id → symbol, blank id, word-boundary token). Made
comprehensive across the same CTC backends via the existing exported functions
(internal dispatch only — no binding changes).

### 2. Whisper `no_speech_prob` on the exception-safe session
The per-segment `<|nospeech|>` posterior existed on the legacy
`whisper_full_get_segment_no_speech_prob` path but was absent from
`crispasr_session_result`. Thread it through in the Whisper branch (sentinel for
non-Whisper backends) and add the C-ABI + Rust accessor.

### 3. Whisper `detected_language` on the exception-safe session
The in-decode acoustic language id (`whisper_full_lang_id_from_state` +
`whisper_lang_str`) was only on the legacy `CrispASR` wrapper. Add
`crispasr_session_detected_language(...)` and `Session::detected_language()`,
mirroring the legacy behavior; non-Whisper backends return the input
language / `"unknown"` (text-LID remains the separate backend-agnostic path).

### 4. Binding mirror
Mirror the two new Whisper accessors across the language bindings
(csharp / go / java / javascript / ruby / python / flutter-dart), keeping the
per-binding lint/test CI green — same rollout pattern as prior accessor work.

## Testing
- Session results carry a real per-segment `no_speech_prob` and a
  `detected_language` matching the legacy `whisper_context` path on a fixture.
- clang-format-18 applied to touched C++; per-binding CI green.

### DIFF
--- a/bindings/csharp/CrispASR/NativeMethods.cs
+++ b/bindings/csharp/CrispASR/NativeMethods.cs
@@ -31,6 +31,12 @@ namespace CrispASR
         internal static extern int crispasr_session_available_backends(
             byte[] outCsv, int outCap);
 
+        // Acoustic detected language (Whisper) as an ISO-639-1 code into outBuf;
+        // other backends fall back to the source-language hint, then "unknown".
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int crispasr_session_detected_language(
+            IntPtr session, byte[] outBuf, int outCap);
+
         // CTC vocabulary access (Omni CTC backend). n_vocab is the piece count;
         // token_text returns a model-owned const char* (do not free) or "" when
         // out of range / unsupported.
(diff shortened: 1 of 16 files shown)


CrispStrobe:
Merged into `main` (`a9f957aa5`) — thank you! Clean, well-scoped additive introspection on the exception-safe session, and the sentinel discipline (`-1.0` / `""` / `"unknown"` for "unavailable") is exactly right.

I reviewed the C++ and confirmed the guards hold: `crispasr_session_token_text` already rejects `id < 0` at the top, so the per-backend upper-bound-only checks are safe; the whisper `whisper_full_lang_id(ctx)` matches the session's `whisper_full(ctx)` path; and the CTC `n_vocab`/blank relationship is consistent (blank in-vocab for wav2vec2, appended for canary).

I added test coverage on top before landing, since the PR shipped only the two live whisper tests:

- **Unit** (`tests/test-session-introspection.cpp`, `[unit]`, no model): null-handle / bad-arg / OOB guards for all four accessors — locks the sentinel contract callers branch on.
- **Live** (`crispasr/tests/integration.rs`): `session_{canary,wav2vec2}_ctc_vocab` exercise `ctc_vocab()` on the backends you newly made comprehensive (only omni-ctc had a vocab test), and `session_ctc_backend_no_speech_sentinel` verifies a real non-whisper session leaves the `-1.0` sentinel + `detected_language` fallback.

Verified 919/919 unit pass and the live CTC vocab/sentinel tests pass against real `wav2vec2-xlsr-en` / `canary-ctc` models. One small conflict against `main`'s package-prep `dart format` in the Flutter binding was resolved (kept your `noSpeechProb`/`detectedLanguage` accessors; `dart analyze` clean).

PR #268, opened 2026-07-18
fix(session): plumb use_gpu into firered-asr open; map firered-asr in GGUF autodetect

oldcai:
Two small firered-asr fixes, found while embedding crispasr-sys in a downstream app that gates FireRed between CPU and Metal:

- `crispasr_session_open_explicit` now sets `p.use_gpu = g_open_use_gpu_tls` for the firered-asr backend, the same way every other backend branch already does. Without it the caller's CPU/GPU choice is silently ignored for firered sessions (the context always opens with the default).
- `crispasr_detect_backend_from_gguf` now maps `general.architecture` `firered-asr`/`firered` to the `firered-asr` backend, so FireRed GGUFs autodetect like the other architectures instead of falling through unrecognized.


CrispStrobe:
Merged to main via ff-merge (rebased, built, unit+live tests pass — 12 pre-existing failures unrelated to firered-asr).

PR #269, opened 2026-07-18
build(sys): disable GGML_NATIVE for cross and x86_64 macOS builds

oldcai:
ggml defaults `GGML_NATIVE=ON` (`-march=native`), which is wrong whenever the compile host is not the machine the binary will run on — and it hard-fails under Rosetta 2: clang's host-CPU probe reports the Apple Silicon die (e.g. `apple-m2`) while targeting x86_64, and cc rejects it with `error: unknown target CPU`. Hit this on an "x86_64" CI runner that is actually Rosetta on Apple Silicon.

This turns it off for any host≠target cross build and for every x86_64 macOS build (a Rosetta toolchain looks like a native x86_64 host, so the cross check alone can't catch it; and Intel-Mac artifacts built for distribution shouldn't be tuned to the build box either). ggml's per-ISA defaults (AVX2/FMA/F16C on x86) still apply, so the result stays on SIMD kernels rather than dropping to scalar.

`CRISPASR_FORCE_GGML_NATIVE=1` opts back in.

### DIFF
--- a/crispasr-sys/build.rs
+++ b/crispasr-sys/build.rs
@@ -198,6 +198,26 @@ fn configure_and_build(src_root: &Path) -> PathBuf {
             .arg("-DCMAKE_CXX_COMPILER_LAUNCHER=ccache");
     }
 
+    // ggml defaults GGML_NATIVE=ON (`-march=native`), which is wrong whenever
+    // the compile host is not the machine the binary will run on — and it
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Merged to main via ff-merge. build.rs-only change — C++ build + 943 unit tests confirmed clean.

PR #270, opened 2026-07-18
build(sys): disable ggml ccache auto-detection

oldcai:
ggml's own cmake auto-detects ccache/sccache as a compiler launcher. Under the Ninja generator, sccache breaks the build on the `GGML_METAL_EMBED_LIBRARY` assembly object: ninja emits depfile rules for the `.s` compile, sccache cannot produce the `.d`, and the build aborts.

build.rs already wires plain ccache explicitly as the launcher when present (with the `CRISPASR_NO_CCACHE` opt-out), and that path is unaffected — so passing `-DGGML_CCACHE=OFF` loses no caching, it just stops ggml from silently picking up sccache and failing on machines that have it on PATH (e.g. Rust CI boxes using `RUSTC_WRAPPER=sccache`).

### DIFF
--- a/crispasr-sys/build.rs
+++ b/crispasr-sys/build.rs
@@ -198,6 +198,13 @@ fn configure_and_build(src_root: &Path) -> PathBuf {
             .arg("-DCMAKE_CXX_COMPILER_LAUNCHER=ccache");
     }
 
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Merged to main via ff-merge. build.rs-only — build + 943 unit tests clean.

PR #271, opened 2026-07-18
build(sys): disable opus/amr file-decode extras

oldcai:
When crispasr-sys drives the cmake build itself (the OUT_DIR fallback path), the optional `.opus`/`.amr` file-decode extras link Homebrew/system dylibs by absolute path (`libopusfile`, `libopencore-amr`, ...). The produced `libcrispasr` then fails to load on any machine that doesn't have those packages installed — a real problem when the crate is embedded in a redistributable app.

Sys-crate consumers feed raw PCM through the C API, so the file-decode extras are unused on this path anyway. This only changes the build.rs-driven configure; the CMake project defaults are untouched for people running cmake directly.

### DIFF
--- a/crispasr-sys/build.rs
+++ b/crispasr-sys/build.rs
@@ -198,6 +198,14 @@ fn configure_and_build(src_root: &Path) -> PathBuf {
             .arg("-DCMAKE_CXX_COMPILER_LAUNCHER=ccache");
     }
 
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Merged to main via ff-merge. build.rs-only — build + 943 unit tests clean.

PR #272, opened 2026-07-18
build: make the shipped library file name configurable

oldcai:
Adds a `CRISPASR_LIB_OUTPUT_NAME` cache variable for the library file name (default stays `crispasr`, so existing builds are unchanged) and has build.rs pass `link_lib_name()` through, so the existing `CRISPASR_LIB_NAME` env var now controls both the `cargo:rustc-link-lib` directive and the file name cmake actually produces.

The C API symbols are unchanged — this just lets downstream bundlers ship the dylib under their own name instead of post-processing/renaming it (and, before this, the rename would break the link since build.rs already supported `CRISPASR_LIB_NAME` but the build always emitted `libcrispasr`).

### DIFF
--- a/crispasr-sys/build.rs
+++ b/crispasr-sys/build.rs
@@ -198,6 +198,10 @@ fn configure_and_build(src_root: &Path) -> PathBuf {
             .arg("-DCMAKE_CXX_COMPILER_LAUNCHER=ccache");
     }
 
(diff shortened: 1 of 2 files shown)


CrispStrobe:
Merged to main via ff-merge. build.rs + CMakeLists.txt — build + 943 unit tests clean, default library name preserved.

PR #277, opened 2026-07-19
fix(audio): preserve unity gain when downsampling

KevinAHM:
## Summary

- use the reduced rational ratio's interpolation factor `L` for polyphase gain compensation
- prevent downsampling from amplifying audio (2x at 48 kHz -> 24 kHz; 1.5x at 24 kHz -> 16 kHz)
- add a focused regression test covering two downsampling ratios and one upsampling ratio

## Why

The filter compensates for zero-stuffing by multiplying by the interpolation factor `L`. The previous code multiplied by `max(L, M)`, which is only equivalent while upsampling. When `M > L`, output amplitude was scaled by `M / L`.

## Validation

- `test-audio-resample [unit]`: 9 assertions pass
- tested 48 -> 24 kHz, 24 -> 16 kHz, and 24 -> 48 kHz DC gain preservation

## AI provenance

The regression analysis, implementation, tests, and PR description were produced by **OpenAI Codex, GPT-5.6-sol (high reasoning)** under @KevinAHM's direction. Codex also executed the local formatting, build, and test validation reported above.

### DIFF
--- a/src/core/audio_resample.cpp
+++ b/src/core/audio_resample.cpp
@@ -122,7 +122,7 @@ std::vector<float> resample_polyphase(const float* in, int n_in, int src_rate, i
         // Compensate for the implicit factor-L upsample (zero-stuffing
         // multiplies the spectrum by 1/L; a corresponding factor-L gain
         // restores unity DC).
-        out[(size_t)i] = (float)(s * (double)up);
+        out[(size_t)i] = (float)(s * (double)L);
     }
     return out;
 }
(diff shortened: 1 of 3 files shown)


CrispStrobe:
Integrated into `main` as 78e573102 (authorship preserved) + follow-up 732288060 which removes the now-dead `up` variable and adds a 48000→44100 regression case (the actual downsample the `--separate` surface exercises). Verified locally: `test-audio-resample` — all 12 assertions pass. Thanks!

PR #279, opened 2026-07-19
fix(webrtc-vad): use native Windows initialization

KevinAHM:
## Summary

- use the standard C++17 `[[noreturn]]` attribute for the WebRTC fatal-message shim
- define `WEBRTC_POSIX` only on non-Windows builds
- allow the vendored WebRTC SPL initialization code to select its existing native `_WIN32` path

## Why

The `webrtc-vad` target unconditionally defined `WEBRTC_POSIX`, which made the vendored `spl_init.c` include `pthread.h` on Windows instead of selecting its existing `CRITICAL_SECTION` implementation. The C++ wrapper also used the GCC/Clang-specific `__attribute__((noreturn))`, causing an earlier MSVC compile failure before the pthread mismatch was reached.

This PR restores the intended platform selection without changing the public API or VAD behavior. POSIX builds continue to define `WEBRTC_POSIX`; Windows builds use the implementation already provided by the vendored WebRTC source.

## Validation

- clang-format 18.1.8 check passes
- focused `webrtc-vad` target compiles and links successfully with MSVC 19.50
- concurrent Windows smoke test completed 400 VAD calls with zero failures and 400 detected segments

## AI provenance

The root-cause analysis, implementation, validation, and PR description were produced by **OpenAI Codex, GPT-5.6-sol (high reasoning)** under @KevinAHM's direction. Codex executed the local formatting, focused Windows build, and concurrent VAD smoke test reported above.


CrispStrobe:
Both changes in this PR are already on `main` (applied while fixing the broken v0.8.14 Windows binaries, with your diagnosis credited in the code comment): the `if(NOT WIN32)` WEBRTC_POSIX guard in src/CMakeLists.txt and the `[[noreturn]]` conversion in src/webrtc_vad.cpp. I verified the WEBRTC_POSIX/spl_init.c pthread-vs-CRITICAL_SECTION reasoning against the vendored source before applying it. Thanks for the precise root-cause — closing as already-integrated.

Issue #254, opened 2026-07-13
OmniVoice GPU


bakamomi:
I tested the `feat/omnivoice-gpu` branch. Simple tts works both on cuda and vulkan. Voice design with tags also seems to work.
1. Voice cloning doesn't work.
2. RTF is 3-4x worse than alternative cpp implementations like https://github.com/ServeurpersoCom/omnivoice.cpp

CrispStrobe:
thank you for testing that work-in-progress branch. it should already work much better now.

bakamomi:
Looks like changes are spread between multiple branches. Here are some tests for the main branch.
```
./crispasr --threads 8 --backend omnivoice-tts -m ../../../models/tts/omnivoice-base-Q8_0.gguf --codec-model ../../../models/tts/omnivoice-tokenizer-Q8_0.gguf --instruct "male, middle aged, low pitch" --tts "CrispASR started as a fork of whisper.cpp and extends that base into a unified speech engine called crispasr, backed by full ggml C++ runtimes for major open-weights ASR and TTS architectures. One build, one binary, one consistent CLI — pick the backend at the command line or let CrispASR auto-detect it from your GGUF file. See Text-to-Speech for the TTS side."
```
This command runs in 14.33 seconds (wall time) and produces 36.4 second audio. RTF = 0.39

```
./build/omnivoice-tts --model ../models/tts/omnivoice-base-Q8_0_cpp.gguf --codec ../models/tts/omnivoice-tokenizer-F32.gguf --instruct "male, middle-aged, low pitch" --lang English -o out.wav < prompt.txt
```
Prompt.txt contains the same line as used for CrispASR.
The command runs in 5.13 seconds and produces 21.61 second audio. RTF = 0.24

CrispStrobe:
Thanks.  Voice cloning is implemented. A corrupt shipped tokenizer is fixed. The RTF gap caused by over-generation should be mostly fixed. Though your box may well differ. **Would you mind re-running your benchmark on current `main`?** The audio length is now apples-to-apples. Residual per-frame gap would be mostly kernel-level.

bakamomi:
The total generated time is now similar.
```
./crispasr --threads 8 --backend omnivoice-tts -m ../../../models/tts/omnivoice-f16.gguf --codec-model ../../../models/tts/omnivoice-tokenizer-f16.gguf --instruct "male, middle aged, low pitch" --tts "CrispASR s
[...]

PR #280, opened 2026-07-19
perf(omnivoice): default codec to GPU on CUDA

KevinAHM:
## Summary

- default the OmniVoice codec to GPU when the main compute backend is CUDA
- retain `CRISPASR_OMNIVOICE_CODEC_GPU=0/1` as an explicit CPU/GPU placement override
- keep the existing CPU default on Metal and CPU backends, and update the environment and OmniVoice documentation

## Why

OmniVoice's CUDA codec implementation already exists, but codec placement remained globally CPU-by-default because the GPU path was slower than CPU-FASTCONV on Apple Metal. That conservative cross-platform default leaves CUDA deployments decoding the DAC on CPU, where decoding can dominate total synthesis time.

CUDA does not share the Metal result. Issue #254 already reported 0.05–0.11 s GPU decode on an RTX 5070 Ti. A fresh same-code A/B on the same GPU class measured approximately 1.4 s on CPU versus 34 ms on CUDA, with PCM cosine 0.99999986.

This PR makes the existing placement policy backend-aware: CUDA selects the GPU codec automatically, while Metal and CPU retain their current behavior. It does not change decoder math, the public API, or non-CUDA defaults.

## Relationship to #254

This completes the CUDA-default follow-up described in the OmniVoice performance plan and confirmed by the reporter's CUDA results in #254. The environment override remains available for regression bisection and unusual systems.

## Validation

- clang-format 18 check passes (`format.sh: OK`, 461 files)
- focused `omnivoice` native target configures, compiles, and links successfully from a clean worktree
- CUDA load validation with the variable unset selected `compute backend = CUDA0` and `codec backend = CUDA0`
- setting `CRISPASR_OMNIVOICE_CODEC_GPU=0` retained `compute backend = CUDA0` and selected `codec backend = CPU`
- 32-step CUDA A/B on the reported PlayerMale voice reduced the 7.52 s segment's decode from 1.25 s to 0.05 s
- CPU/GPU final outputs had identical length, PCM cosine 0.999999862, RMSE 0.0000352, and max absolute difference 0.000458

## AI provenance

The source/history review, implementation, validation, and PR description were produced by **OpenAI Codex, GPT-5.6-sol (high reasoning)** under @KevinAHM's direction. Codex executed the local formatting, isolated build, CUDA backend-selection checks, performance benchmark, and PCM comparison reported above.


CrispStrobe:
Integrated into `main` as dbf211a0b (authorship preserved). Confirmed it matches the proven `unified_cfg` CUDA-default idiom; Metal/CPU keep the M1-verified FASTCONV path; env-revertible via CRISPASR_OMNIVOICE_CODEC_GPU=0. Builds clean. Thanks!

PR #281, opened 2026-07-19
fix(omnivoice): honor TTS seed controls

KevinAHM:
## Summary

- pass the configured session-open seed into the OmniVoice context
- route live `crispasr_session_set_tts_seed` calls to OmniVoice
- expose a small backend setter so subsequent syntheses use the requested seed

## Why

OmniVoice already samples from a generator-local `std::mt19937`, but the C API never forwarded either the session-open seed or later `set_tts_seed` calls to that generator. The setter therefore returned the unsupported-backend result and different requested seeds produced the same output.

This PR connects the existing seed controls without changing the public API or the server/Python layers.

## Validation

- focused `omnivoice` target and `crispasr_c_api.cpp` translation unit compile successfully
- all 26 existing session-setter tests pass
- end-to-end CUDA validation on one live session with seeds `111`, `222`, then `111` produced byte-identical codes for the repeated seed, while the two different seeds differed at 97.45% of code positions

## AI provenance

The root-cause analysis, implementation, validation, and PR description were produced by **OpenAI Codex, GPT-5.6-sol (high reasoning)** under @KevinAHM's direction. Codex executed the local formatting, build, session-setter tests, and end-to-end seed validation reported above.

### DIFF
--- a/src/crispasr_c_api.cpp
+++ b/src/crispasr_c_api.cpp
@@ -2644,6 +2644,7 @@ CA_EXPORT crispasr_session* crispasr_session_open_explicit(const char* model_pat
         p.verbosity = g_open_verbosity_tls;
         p.use_gpu = g_open_use_gpu_tls;
         p.flash_attn = g_open_flash_attn_tls;
+        p.seed = g_open_seed_tls;
         s->omnivoice_ctx = omnivoice_init_from_file(model_path, p);
         if (!s->omnivoice_ctx) {
             delete s;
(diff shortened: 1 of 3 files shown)


CrispStrobe:
Integrated into `main` as e907acbcf (authorship preserved) + follow-up 773ebd11e adding a live seed-determinism regression test (tests/test-omnivoice-seed.cpp): same seed → byte-identical codes, different seed → different, at the fast text→codes stage. Verified the setter routing (null-handle → -1) still passes and the test is SKIP-clean without a model. Thanks!

Issue #302, opened 2026-07-24
CUDA backend for Pascal


subof:
In fact, this does not work on the Quadro P5000. Is there a way to support this?
I'm using VULKAN now, but things aren't going smoothly there either.

CosyVoice3 (CrispASR) — the crispasr server crashed during synthesis.
Server log:
tok=150,T_mel=300) fleurs-fr(T_tok=126,T_mel=252) fleurs-es(T_tok=141,T_mel=282) fleurs-ko(T_tok=120,T_mel=240)
crispasr[cosyvoice3-tts]: 8 voice(s) available: zero_shot fleurs-en fleurs-de fleurs-zh fleurs-ja fleurs-fr fleurs-es fleurs-ko
crispasr-server: warmup completed in 0 ms
crispasr-server: backend 'cosyvoice3-tts' loaded, model 'C:\Users\User\AppData\Roaming\Subtitle Edit\CrispASR\models\cosyvoice3-llm-q4_k.gguf'
crispasr-server: auto-enabling punctuation restoration for backend 'cosyvoice3-tts'
crispasr[punc]: using cached C:\Users\Uder/.cache/crispasr/fireredpunc-q4_k.gguf
fireredpunc: 12L, d=768, ffn=3072, heads=12, vocab=21128, labels=5
crispasr-server: loaded punctuation model 'C:\Users\User/.cache/crispasr/fireredpunc-q4_k.gguf'

crispasr-server: listening on 127.0.0.1:6729
  POST /inference                  вЂ” upload audio (native JSON)
  POST /v1/audio/transcriptions    вЂ” OpenAI-compatible API
  POST /v1/audio/speech            вЂ” TTS (OpenAI-compatible)
  POST /load                       вЂ” hot-swap model
  GET  /health                     вЂ” server status
  GET  /backends                   вЂ” list backends
  GET  /v1/models                  вЂ” model info
  GET  /v1/voices                  вЂ” list voices in --voice-dir
       /voices, /v1/audio/voices   вЂ” aliases (llama-swap compat)
  POST /v1/voices                  вЂ” upload voice file (multipart)
  DELETE /v1/voices/:name          вЂ” delete voice file

cosyvoice3_tts:campplus loaded 617 tensors
cosyvoice3_tts:s3tok loaded 198 tensors  blocks=12 d=1280 h=20 kernel=31 codebook=6561
ggml_cuda_graph_set_enabled: disabling CUDA graphs due to GPU architecture
[...]

PR #307, opened 2026-07-25
ci(windows): add Tesla P100 to CUDA release targets

W1nge:
## Summary

Add Tesla P100 (`sm_60`) SASS to the two existing Windows CUDA 12.8
release artifacts:

- `crispasr-windows-x86_64-cuda.zip`
- `libcrispasr-windows-x86_64-cuda.tar.gz`

Also document the equivalent local Windows build command.

## Problem

Tesla P100 is a Pascal GPU with compute capability 6.0. CrispASR already
builds and runs on it when `CMAKE_CUDA_ARCHITECTURES=60` is set, but the
official Windows CUDA artifacts currently start at `sm_75`. Loading those
artifacts on a P100 can therefore fail with:

```text
no kernel image is available for execution on the device
```

## Changes

- Prepend `60-real` to the architecture list in the Windows CUDA CLI release
  job.
- Prepend `60-real` to the architecture list in the Windows CUDA shared-library
  release job.
- Add a P100 example to the Windows build documentation.
- Document that Pascal builds require CUDA 12.x because CUDA 13 no longer
  generates Pascal code.

Artifact names and public APIs are unchanged. Linux, CUDA 13, CPU, and Vulkan
release jobs are intentionally unchanged.

## Validation

- The modified GitHub Actions workflow parses successfully as YAML.
- `git diff --check` passes.
- Only the two Windows CUDA 12.8 jobs contain the new `60-real` target.
- A complete Windows CUDA 12.4/Ninja build with
  `CMAKE_CUDA_ARCHITECTURES=60` generated:

  ```text
  --generate-code=arch=compute_60,code=[compute_60,sm_60]
  ```

- That build produced `ggml-cuda.dll` and was validated on a
  Tesla P100-PCIE-16GB reported as compute capability 6.0. The MiMo ASR worker
  initialized on CUDA and completed long-audio transcription tests.

## Release size impact

This deliberately adds another real CUDA architecture to the existing
artifacts, so build time and archive size will increase. For reference, the
v0.8.22 Windows CUDA CLI and shared-library artifacts are approximately
691 MB and 810 MB respectively.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1367,7 +1367,7 @@ jobs:
           fail_on_unmatched_files: false
 
   # ── Windows x86_64 shared libs + headers (CUDA) ──────────────────────────────
-  # CUDA 12.8 toolkit (Blackwell-aware). Unlike the Linux CUDA tarball,
+  # CUDA 12.8 toolkit (Tesla P100 through Blackwell). Unlike the Linux CUDA tarball,
   # we bundle the CUDA runtime DLLs (cudart / cublas / cublasLt) so the
   # tarball is self-contained for downstream Windows consumers — matches
   # the existing build-windows-cuda EXE bundle's pattern.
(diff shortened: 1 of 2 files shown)


CrispStrobe:
Thanks @W1nge — this was exactly the right catch, and the note that **CUDA 13 no longer generates Pascal code** flagged a real bug we had just introduced in #302 (we had added `61-real;70-real` to the CUDA-13 build jobs, which `nvcc 13` would have rejected with `Unsupported gpu architecture`).

Your P100/`sm_60` work has been incorporated and extended in `018bbac4` on `main`:

- **`60-real` (P100)** is now prepended to **all five** CUDA-12.x build jobs (Linux CLI, Linux+Windows shared-libs, Windows CLI, and the `main-cuda-12` Docker image) — not only the two Windows jobs, so P100 gets native SASS everywhere CUDA 12.x ships it.
- The **CUDA-13** jobs (`build-linux-x86_64-cuda13`, `main-cuda`, `main-cuda-portable`) floor at `sm_75`, since `nvcc 13` dropped Maxwell/Pascal/Volta — Pascal/Volta users are served by the CUDA-12.x artifacts.
- Your `docs/install.md` P100 note is included verbatim.

You are credited via `Co-authored-by` on the commit. Closing since it is fully incorporated — ships in the next release. 🙏

PR #343, opened 2026-08-11
docs: add TOC and section links to tts.md

Juste-Leo2:
### Description:
This PR adds a table of contents for tts.md, as well as clickable links in the table of the different TTS backends.
I noticed this when I wanted to use Qwen3-TTS and had to use Ctrl+G in the docs :)

### Additional notes:
- This PR is inspired by what was done in cli.md
- A dead link was also removed (in cli.md)
- A verification script was used for the links

### Note on AI usage:
I used OpenCode to make the changes


CrispStrobe:
Merged as `8da6a531` (resolved and pushed locally, so GitHub did not auto-close this). Thanks — a TOC on a file this long is a real usability win, and the anchor work is correct.

Verified independently rather than from the description, since anchors are exactly the thing that rots silently:

- The dead link you removed is genuinely dead — `docs/cli.md` pointed at `tts.md#timing-quality-tada_num_candidates` while the heading slugs to `#timing-quality-crispasr_tada_num_candidates`.
- After the merge, **all 79 internal anchors across `tts.md` and `cli.md` resolve** (33 before).
- The TOC covers **all 20** `##` sections, none missed.
- The rows you left unlinked are the ones in secondary tables (cross-lingual effects, env switches) rather than the backend catalogue — correct scope.

**One conflict worth knowing about.** `main` had corrected two licence strings since you branched: `outetts` to CC-BY-NC-SA-4.0 (non-commercial, ShareAlike) and `pocket-tts` to CC-BY-4.0 plus gated-use conditions. Your branch still carried the older text on those two rows. I took `main`'s licence text with your links — licence strings are compliance-relevant here, so the newer ones win. Nothing wrong on your side; the rows just moved under you.

Optional follow-up if you feel like it: `omnivoice` in the main table is still unlinked, though there is an "OmniVoice: ISO 639-3 ids" section it could point at.

Thanks for the contribution, and for saying up front which tool you used.

PR #345, opened 2026-08-12
feat: expose MOSS valid-frame metadata in stable C ABI (#344)

### LINKED ISSUE #344: MOSS-Music stage extraction: expose valid-frame metadata in stable C ABI
## Context

The `moss_audio` backend already exposes encoder, DeepStack tap, and adapter stage helpers in `src/moss_audio.h` / `src/moss_audio.cpp`. A downstream MOSS-Music-8B-Thinking feature pipeline needs a stable way to obtain the final 1280-D encoder stream, taps 8/16/24, and the 4096-D adapted stream together with observed valid-frame metadata, without the 3000-frame mel padding being mistaken for real content.

This request is only for a stable source-level C ABI. Downstream HTTP, CAS publishing, authentication, and rights policy remain outside CrispASR.

## Acceptance criteria

- [ ] Add new exported C symbols without changing existing `moss_audio_*` signatures. A metadata-returning encoder entrypoint returns the existing tensor and `out_T_enc` / `out_d`, plus caller-allocated per-chunk valid-frame counts, `out_num_chunks`, the pre-pad `out_T_mel_actual`, and `out_total_valid`.
- [ ] Reuse the existing encoder chunk loop and its computed `valid_lens`; do not derive valid frame counts from a global floor formula.
- [ ] Preserve the existing final 1280-D encoder output and DeepStack taps 8/16/24. Report the adapter output dimension from GGUF metadata/weights rather than hardcoding it.
- [ ] Correct the stale header comment: `moss_audio.llm.hidden_size` is 2560 for MOSS-Audio-4B and 4096 for MOSS-Music-8B-Thinking.
- [ ] Fail closed at load time if `llm_hidden` disagrees with the adapter output-row shape; do not silently accept a 2560 fallback for a 4096-D checkpoint.
- [ ] Document ownership and threading: one context has one in-flight stage run; callbacks, if added, run synchronously on the caller thread and must not re-enter the same context; caller owns returned allocations/copies.
- [ ] Preserve cached-encoder-graph invalidation/rebuild behavior on the new entrypoint.
- [ ] Add a differential test proving the sum of per-chunk valid frames equals `out_total_valid` and that a sub-30-second input does not expose zero-content frames created only by the 3000-frame mel pad.

## Suggested smallest patch

- `src/moss_audio.h`: declare a metadata-returning encoder API and document dimensions, ownership, and single-inflight behavior.
- `src/moss_audio.cpp`: expose the existing per-chunk `valid_lens`, retain the pre-pad mel length, validate adapter/LLM dimensions, and keep existing symbols untouched.

An optional chunk-local callback may follow the existing `moss_audio_process_cb` pattern, but it is not required if the metadata-returning entrypoint can return final/tap/adapter tensors without duplicate model execution.

jajmangold:
## Summary

Adds an additive, backward-compatible C ABI for truthful MOSS audio valid-frame metadata:

- `moss_audio_compute_mel_meta` reports the observed pre-padding mel-frame count while preserving the existing padded mel-buffer contract.
- `moss_audio_plan_chunks` exposes exact three-stage stride-2 chunk planning and per-chunk valid encoder-frame counts.
- `moss_audio_run_encoder_meta` excludes synthetic pad frames, returns exact valid counts, and preserves byte-identical behavior when `T_mel_actual == T_mel`.
- Existing `moss_audio_compute_mel`, `moss_audio_run_encoder`, and `moss_audio_run_adapter` signatures remain unchanged.
- Adapter output width is derived from the loaded GGUF tensor and checked against `llm.hidden_size` at load time.
- On any NULL encoder return, caller `ds_tap_*` output slots remain untouched; internal failure paths cannot publish freed tap buffers.

## Validation

- Release build completed successfully.
- Hermetic metadata suite: **4 cases / 220 assertions pass**.
- The tap failure-contract regression was verified to fail before commit `c38739d` and pass after it.
- Existing MOSS audio hermetic tests pass.
- Source/header formatting is clean for CI-gated paths.
- Independent adversarial re-review approved this draft after checking pre-pad semantics, chunk/stride math, buffer bounds, ownership, adapter dimensions against real MOSS-Music GGUF geometry, ABI linkage, and failure paths.
- A model-backed differential test is included and skips cleanly without `CRISPASR_MODEL_MOSS_AUDIO`; it should run with the governed MOSS audio model before final merge.

A repository-wide CTest run also exposed unrelated environment/live-test failures (missing optional models, `/mnt/storage` permissions, absent `python` shim, and live server/model tests); the focused changed surface is green.

## Risk and rollback

Risk is limited to the additive MOSS audio ABI and its shared encoder implementation. The legacy entrypoints retain their signatures and success behavior. The strict adapter-dimension check intentionally refuses mislabeled/inconsistent GGUFs. Failure paths are safer because tap outputs are not published on NULL return.

Rollback is the two commits in this branch; no data migration or model artifact change is involved.

Closes #344

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -18,6 +18,18 @@ Trace and fix the HIP-vs-CPU talker divergence, add a regression guard for the
 first sampled token / normal termination, and validate the real GPU path when
 available.
 
+## CLAIMED 2026-08-11 — #344 MOSS valid-frame metadata in stable C ABI
+
+Worktree: `worktrees/crispasr-344-moss-valid-frames`
+(branch `feat/344-moss-valid-frame-metadata`).
+Additive MOSS encoder/tap/adapter valid-frame metadata for the downstream
+MOSS-Music-8B-Thinking feature pipeline: `moss_audio_plan_chunks`,
+`moss_audio_compute_mel_meta` (preserves pre-pad `T_mel_actual` — never inferred
+from padded zeros/floor), and `moss_audio_run_encoder_meta` (existing chunk loop
+reused, caller-allocated per-chunk valid counts, adapter output dim reported from
+GGUF weights, fail-closed llm_hidden vs adapter-row check). Existing
+`moss_audio_*` symbols unchanged. Hermetic CPU test + live differential test.
+
 ## LANDED 2026-08-10 — voxtral-tts pre-tokenizer parity (c69ac61b, from #338)
 
 Carried out of #338 after it closed. Reporter measured 5/57 vs `mistral-common`;
--- a/src/moss_audio.cpp
+++ b/src/moss_audio.cpp
@@ -408,6 +408,23 @@ static bool moss_audio_load_model(moss_audio_model& model, moss_audio_vocab& voc
     model.adapter.up_w = require(model, "adapter.up.weight");
     model.adapter.down_w = require(model, "adapter.down.weight");
 
+    // Issue #344: the adapter's output-row shape comes from the weights, not a
+    // hardcoded 2560. down_w is (adapter_hidden, llm_hidden) in ggml ne terms,
+    // so ne[1] is the output feature count. Fail closed if the GGUF kv
+    // moss_audio.llm.hidden_size disagrees — never silently accept a 2560
+    // fallback for a 4096-D MOSS-Music-8B-Thinking checkpoint.
+    {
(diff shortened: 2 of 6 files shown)


CrispStrobe:
Superseded by #349, which includes this PR plus the validation fixes and tests.

Issue #350, opened 2026-08-13
parakeet-tdt-0.6b-v3: v0.8.24 unified dispatch regressed long-form chunking — 94 % → 66 % transcript coverage vs v0.8.9, whole spans silently dropped


davideme:
## Summary

With **parakeet-tdt-0.6b-v3** (multilingual, q4_k), a session call that asks for chunked long-form
transcription is routed to a **single full-length decode** as of **v0.8.24**, and silently loses large
contiguous spans of speech — **45 s in one gap** on a 230.7 s file, 34 % of the transcript missing.

**v0.8.9 does not have this problem.** Same file, same model, same call, no env vars:

| CrispASR | words | coverage vs reference | longest dropped run |
|---|---|---|---|
| **0.8.9** (`c02d9200`) | **430** | **94 %** | 3 words |
| **0.8.24** (`5ebe79f6`) | 291 | 66 % | 45 words |

Two things combine to cause it:

1. `crispasr_c_api.cpp:5079` — the "unified dispatch" (default ON, new in v0.8.24) maps a session call
   with `chunk_seconds = 0` to `chunk_seconds_explicit = false`, then returns early. That bypasses
   `use_single_pass`, the `CRISPASR_PARAKEET_CHUNK_*` env vars **and**
   `parakeet_session_chunked_merge`. In v0.8.9 that same call reached the merge path (20 s windows /
   8 s overlap → 19 per-window decodes).
2. `parakeet_orchestrate.cpp:234` — `int stream_threshold_s = is_ja ? 12 : 300;`
   Having landed in the orchestrator, anything below 300 s gets `SINGLE_PASS`. But the decoder is
   documented (and observed) to drop sections past ~30 s, so 30–300 s is a dead zone: too long to
   decode reliably, too short to reach `LONGFORM`.

`0` is the documented "use per-model defaults" value for
`crispasr_session_transcribe_chunked[_lang]`, so callers explicitly asking for chunked long-form
transcription are the ones silently getting an unchunked single pass.

> **Scope of these measurements.** Everything below was measured with
> **parakeet-tdt-0.6b-v3-q4_k** only. Both code paths are keyed on `is_ja`, so on inspection they
> apply to every non-Japanese Parakeet model — but I have not tested any other model, and the
> ~30 s degradation point in particular may well be model-specific. Treat the non-JA generalisation
> as inferred from the source, not measured.

## Environment

```
[...]

PR #353, opened 2026-08-13
perf(parakeet): 2.1x long-form throughput — at or above FluidAudio GPU mode

davideme:
## Why

Benchmark the parakeet backend against [FluidAudio](https://github.com/FluidInference/FluidAudio), a Swift/CoreML on-device ASR SDK running the same NVIDIA Parakeet TDT 0.6B v3 model.

The short-utterance path already was already on par. Long-form was not: it ran at **25.6× RTFx, well below our own short-utterance rate of ~40×**, which is backwards, longer audio should amortise cost, not lose it. This PR fixes that. Long-form now sustains **~59× RTFx**, above FluidAudio's GPU-mode figure, and the same-model accuracy lead widens rather than being traded away.

## Result

All numbers: Apple M1 Pro (16 GB), macOS 26.5, `parakeet-tdt-0.6b-v3` Q4_K (CrispASR) vs the CoreML int8 encoder (FluidAudio), identical audio, WER computed by one normalizer over both engines' output.

**Short utterances** — 300 LibriSpeech test-clean, 44 min, 37 speakers. *(This path is untouched by the PR; included to establish the GPU-mode baseline.)*

| engine | encoder | RTFx | WER |
|---|---|---|---|
| **CrispASR** | **Metal** | **40.1×** | **1.76 %** |
| FluidAudio | GPU | 38.9× | 2.72 % |
| FluidAudio | CPU | 35.2× | 2.71 % |
| FluidAudio | ANE | 59.9× | 2.78 % |

At the same execution tier we are slightly faster and **1.5× more accurate**. 

**Long-form** — two ~12 min LibriSpeech concatenations. This is what the PR changes.

| engine | config | A: RTFx | A: WER | B: RTFx | B: WER |
|---|---|---|---|---|---|
| **CrispASR (this PR)** | **Metal, default** | **58.9×** | **1.87 %** | **59.1×** | **1.38 %** |
| CrispASR (this PR) | Metal, `--vad` | 51.2× | 1.87 % | 47.1× | 1.29 % |
| CrispASR (before) | Metal, pre-PR | 25.6× | 1.87 % | 27.9× | 11.68 % ✗ |
| FluidAudio | ANE, default | 114.2× | 7.14 % ✗ | 98.9× | 6.11 % ✗ |
| FluidAudio | ANE, `--no-mel-context` | 182.7× | 2.21 % | 134.4× | 3.13 % ✗ |

**2.1–2.3× throughput, WER unchanged or far better.** Long-form went from 0.64× our short-utterance rate to 1.47× — it is no longer the weak spot. On File B the old path was also silently dropping 233 of 2175 words; the new one drops none.

("before" measured on this same build with `CRISPASR_PARAKEET_LONGFORM_WINDOW=300 CRISPASR_PARAKEET_PIPELINE=0`, which reproduces the pre-PR behaviour exactly.)

Note: FluidAudio only exposes `--encoder-compute-units` on its `asr-benchmark` command (short utterances), so its long-form columns are its ANE default — i.e. the long-form comparison is against a *faster* tier than GPU, not a like-for-like one.

## What changed

Four commits, each independently built and behaviour-tested.

**1. `perf(parakeet): decouple longform window from single-pass cap`**

`stream_threshold_s` was doing two jobs: the cap below which audio gets one exact pass, and the size of each LONGFORM piece. Tuning one moved the other.

The FastConformer encoder materialises an O(T²) relative-position bias, so covering the same audio in k windows costs ~T²/k. Measured 300 s → 90 s windows: **1.85× on Metal, 1.53× on CPU** — the direction is architectural and hardware-independent. The old 300 s window also silently dropped 233 of 2175 words on File B (11.68 % WER vs 1.38 %); the codebase already documented the TDT decoder losing track around 30–60 s, so 300 was ten times past a known limit.

But lowering the shared value pushed 90–300 s audio off the seamless single-pass path (127 s file: 0.91 % → 1.21 % WER). So the cap stays **300** and a new `kParakeetLongformWindowS` (**90**) sizes LONGFORM pieces. Effective window is `min(window, cap)`, so an existing `CRISPASR_PARAKEET_STREAM_THRESHOLD=N` override still shrinks windows exactly as before for N ≤ 90. New `CRISPASR_PARAKEET_LONGFORM_WINDOW` sets the window directly.

**2. `perf(parakeet): pipeline encode and decode across longform windows`**

`transcribe_longform` ran mel → encoder → TDT decode strictly in sequence per window, and started window N+1 only once N finished. Encoder is GPU, decoder is CPU cblas — so one processor was always idle: measured **0.42 cores**, with stage sum equal to wall time (encoder 67 %, decode 32 %).

The windows are independent, so nothing forced it. Window boundaries are now planned up front, and a producer thread encodes window N+1 while the caller decodes window N. **18.2 s → 11.3 s (1.49×), byte-identical output.**

Safety: the threads touch disjoint `parakeet_context` state. That stops being true when the decoder is itself a ggml graph on `ctx->backend` (CUDA/Vulkan default), so `parakeet_decode_uses_backend()` is exported and pipelining disables itself there. `CRISPASR_PARAKEET_PIPELINE=0/1` forces it.

**3. `perf(cli): pipeline encode and decode across VAD slices`**

The dispatcher's slice loop had the same serialisation in a different place. Adds an **opt-in** split-transcribe pair to `CrispasrBackend` (default-off, so no other backend is affected), implemented for parakeet. One model resident, versus the N copies a `-p` worker pool needs. **19.8 s → 13.6 s, byte-identical.**

**4. `chore(parakeet): report BENCH/DEBUG stages on the split path`**

`CRISPASR_PARAKEET_BENCH`/`DEBUG` silently under-reported once long-form stopped going through `parakeet_transcribe_ex`. Also rewords the long-audio CLI hint, which recommended `--vad` as the fast option for long files — no longer true for parakeet (11.2 min file: default 11.2 s vs `--vad` 13.1 s). `--vad` remains the accuracy-robust choice and the way to skip silence, so the hint now says that.

## Verification

- **Output identity:** pipelined vs serial **byte-identical** on both long files and the VAD path; mid-band 127 s output byte-identical to the pre-PR single-pass result; `long.wav` hash matches the pre-PR reference.
- **Determinism:** 5 repeated pipelined runs + serial reference → 1 distinct output (no race).
- **Regression:** 300-utterance set WER **1.76 %**, unchanged.
- `test-parakeet-strategy`: **21/21 assertions pass**.
- `samples/jfk.wav` correct with and without `--vad`.
- Env semantics: default 8 windows; `STREAM_THRESHOLD=60/30` → 12/25 windows (unchanged from before the split); `LONGFORM_WINDOW=300` → clamped to cap.

## Risks and limits

- **Accuracy vs window size is content-dependent and non-monotonic.** File B degrades at 300 **and** 150 **and** 60 s — there is no single "cliff below N". 90 was best-or-near-best on both files tested, but it is a robust empirical pick, not a derived optimum. The header records the full table and says so; re-measure on your own material before assuming it.
- Benchmarks are **two files, both concatenated LibriSpeech read English, one machine.** The speed result is architectural and reproduced on Metal and CPU; the accuracy column is the part that may not transfer.
- Pipelining is **auto-disabled** on CUDA/Vulkan (ggml decode) and when `--return-logits` is set. Not exercised on those backends here — worth a look from someone with the hardware.
- New threads are created per long-form run and per VAD slice-run. Bounded queue (2 encoder buffers, ~4.5 MB each at 90 s), joined before anything else touches the model.

## Reproducing

FluidAudio built from source at HEAD; both engines run on the same 300 LibriSpeech test-clean utterances and the same two concatenations. WER is aggregate (corpus-level), computed over both engines' raw output with one normalizer — it will not match FluidAudio's self-reported per-utterance mean.

### DIFF
--- a/crisp_lid/src/text_lid_dispatch.cpp
+++ b/crisp_lid/src/text_lid_dispatch.cpp
@@ -231,8 +231,7 @@ std::string text_lid_resolve_path(const std::string& arg, const std::string& cac
             fprintf(stderr, "text_lid: backend '%s' not in registry\n", backend);
             return "";
         }
-        if (crispasr_license_requires_acceptance(entry.license) &&
-            !crispasr_license_accepted(entry.license, "")) {
+        if (crispasr_license_requires_acceptance(entry.license) && !crispasr_license_accepted(entry.license, "")) {
             fprintf(stderr, "text_lid: refusing restricted model '%s' without --accept-license %s\n",
                     entry.filename.c_str(), crispasr_license_tag(entry.license).c_str());
             return "";
@@ -249,8 +248,7 @@ std::string text_lid_resolve_path(const std::string& arg, const std::string& cac
     //    without first downloading. Same fallback the main `-m` path uses.
     CrispasrRegistryEntry entry;
     if (crispasr_registry_lookup_by_filename(basename_of(arg), entry)) {
-        if (crispasr_license_requires_acceptance(entry.license) &&
-            !crispasr_license_accepted(entry.license, "")) {
+        if (crispasr_license_requires_acceptance(entry.license) && !crispasr_license_accepted(entry.license, "")) {
             fprintf(stderr, "text_lid: refusing restricted model '%s' without --accept-license %s\n",
                     entry.filename.c_str(), crispasr_license_tag(entry.license).c_str());
             return "";
--- a/examples/cli/crispasr_backend.h
+++ b/examples/cli/crispasr_backend.h
@@ -178,6 +178,40 @@ public:
     // until the next transcribe() / shutdown().
     virtual const crispasr_ctc_logits* last_ctc_logits() const { return nullptr; }
 
+    // ---- Optional split transcribe (encode ∥ decode pipelining) ----
+    //
+    // Encoder-decoder backends typically run the encoder on the GPU and the
+    // decoder on the CPU. Processing N slices with transcribe() serialises the
+    // two, leaving one processor idle at all times. A backend that can hand out
+    // its intermediate encoder state lets the caller overlap the encode of
+    // slice N+1 with the decode of slice N — the same throughput win as a
+    // worker pool, without the N-times model memory.
+    //
+    // Contract: encode_slice() may be called on a WORKER thread while
+    // decode_slice() runs on another, so a backend may only advertise this when
+    // its encode and decode paths touch disjoint state. Calls are issued in
+    // slice order and decodes happen in the same order. The handle is opaque
+    // and must be consumed by exactly one decode_slice(), which frees it;
+    // release_encoded() frees an unconsumed handle on an error path.
+    struct encoded_slice {
+        void* h = nullptr; // backend-owned; null = encode failed
+    };
+    virtual bool supports_split_transcribe() const { return false; }
+    // Whether THIS slice length can go through encode_slice/decode_slice. A
+    // backend may route long inputs through a multi-window path that the split
+    // pair does not reproduce. The caller must check every slice up front and
+    // fall back to the fully sequential path if any slice says no — mixing
+    // transcribe() into a running pipeline would encode on two threads at once.
+    virtual bool can_split_slice(int /*n_samples*/, const whisper_params& /*params*/) const { return false; }
+    virtual encoded_slice encode_slice(const float* /*samples*/, int /*n_samples*/, const whisper_params& /*params*/) {
+        return {};
+    }
+    virtual std::vector<crispasr_segment> decode_slice(encoded_slice /*enc*/, int64_t /*t_offset_cs*/,
+                                                       const whisper_params& /*params*/) {
+        return {};
+    }
+    virtual void release_encoded(encoded_slice /*enc*/) {}
+
     // Optional stereo-aware overload for backends that can split stereo
     // channels for diarization (currently: whisper). Default
     // implementation falls through to mono transcribe(); override when
--- a/examples/cli/crispasr_backend_parakeet.cpp
+++ b/examples/cli/crispasr_backend_parakeet.cpp
@@ -166,6 +166,76 @@ public:
         return out;
     }
 
+    // ---- Split transcribe: encode ∥ decode across dispatcher slices ----
+    //
+    // Used by the VAD / chunk slice loop, where each slice is short enough to be
+    // exactly one encode + one decode — i.e. transcribe()'s SINGLE_PASS path cut
+    // in half. The encoder runs on ctx_'s ggml backend (GPU) and the TDT decoder
+    // on the CPU via cblas, so the caller can overlap them.
+    //
+    // Only advertised when the decoder really is the CPU path: on CUDA/Vulkan
+    // parakeet_decode_frames() is itself a ggml graph on ctx_'s backend and
+    // would race the encoder. JA models are excluded because they never take the
+    // plain single-pass route (#89).
+    struct enc_state {
+        float* buf = nullptr;
+        int T_enc = 0;
+        int d_model = 0;
+    };
+
+    bool supports_split_transcribe() const override {
+        return ctx_ != nullptr && !is_ja_model_ && parakeet_decode_uses_backend(ctx_) == 0;
+    }
+
+    // A slice qualifies only when transcribe() would run it as ONE pass. A
+    // longer slice takes the LONGFORM/STREAMED multi-window route, whose
+    // per-window context and merging this split path does not reproduce.
+    bool can_split_slice(int n_samples, const whisper_params& params) const override {
+        if (!ctx_ || n_samples <= 0)
+            return false;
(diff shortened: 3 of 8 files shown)


CrispStrobe:
Merged as `26799e3a` — thanks, this is a well-built PR. The env gates on both pipelines, the fixpoint reasoning, the honest "risks and limits" section and the measurement table are all above the usual bar, and the `kParakeetLongformWindowS` split (window ≠ single-pass cap) is exactly the right decomposition.

The threading audit holds up: `parakeet_context` really does split cleanly — encoder owns `sched`/`compute_meta`/`cached_enc_*`, decoder owns `pred_w`/`joint_w`, `model`/`vocab` read-only — and `parakeet_init_ggml_decoder` genuinely touches nothing when `ggml_dec` is false, so the `parakeet_decode_uses_backend()` gate is sound. I reproduced your byte-identity result exactly on defaults.

Three things needed fixing on top, in `e33eba8c`. They're all the same root cause: **the split pair is not equivalent to `transcribe()`** — `transcribe()` is prologue + encode + decode + epilogue, and the split reproduced the middle two.

**1. A crash — concurrent encode on one ggml scheduler.**

`finish_slice()` runs the #89 gap-fill whenever `vad_slice_cap_seconds() > 0`, and that pass calls back into `be.transcribe()`. On the pipelined path that runs on the **consumer** thread while the producer is encoding. Nothing gated on it:

```
CRISPASR_PARAKEET_VAD_SLICE_CAP=12 crispasr --backend parakeet \
  -m parakeet-tdt-0.6b-v3-q4_k.gguf -f fleurs_300s.wav --vad --verbose

crispasr: pipelining encode/decode over 31/31 slices
.../ggml-backend.cpp:478: GGML_ASSERT(ggml_are_same_layout(src, dst) && "cannot copy tensors with different layouts") failed
rc=134 (SIGABRT)
```

The identical command with `CRISPASR_SLICE_PIPELINE=0` returns rc=0 and 614 words. It's the hazard your own comment names — "the pipeline must never fall back to `transcribe()` while the producer is running" — just arriving from the dispatcher's post-pass rather than from a non-qualifying slice. Non-JA parakeet has a 0 cap by default, so it takes that env var to reach; it's a supported knob though.

Related: the `CRISPASR_SLICE_PIPELINE` override re-derived the decision from a shorter list than the default expression, so `=1` could switch `return_logits` and the worker-pool condition back on. Both holes are the same shape — a safety condition spread across two expressions loses a term. They're now one pure predicate (`examples/cli/crispasr_split_pipeline.h`) where the env var can only turn the pipeline *off*, with a unit case per condition, red-verified against the old semantics.

**2. Five CLI flags silently dropped.**

`transcribe()` applies sticky per-call state to the context on every call — `set_temperature`, `set_beam_size`, `set_att_context`, `set_maes`, `set_hotwords` — and those setters exist nowhere else in the CLI; `init()` sets none of them. So with `--vad --beam-size 4`:

```
pipelined: "Emergency Management Office said that there were no damages..."
serial:    "The emergency management office said that there were no damages..."
```

The pipeline decoded greedily. Applying them from `encode_slice` would mutate decode state on the producer thread, so they now go through a `begin_split_run()` hook the dispatcher calls once, on its own thread, before any producer starts.

**3. Issue #350's dropped-span repair stopped running.**

`parakeet_decode_frames_to_segments()`'s comment says "same tail as the SINGLE_PASS branch" — but that branch ends with `if (repair) gap_fill_segments(...)`, and `repair = !is_ja` is always true here since `supports_split_transcribe()` requires non-JA. With the repair reachable (`CRISPASR_GAP_FILL_MIN_CS=50`): serial recovers 2 words, pipelined recovers 0. Not a niche path — `--verbose` reports `pipelining encode/decode over 26/26 slices` on a plain `--vad` run.

It re-encodes, so it can't live inside `decode_slice`. It's exposed as `parakeet_repair_segments()` and driven by a `repair_slice()` hook the dispatcher calls after the pipeline joins and before `finish_slice` trims — where `transcribe()` applies it. Segments are held until the join; progress is ticked at decode time so a pipelined run still reports incrementally.

Also added an RAII join guard: an exception escaping the consumer loop reached `~std::thread` with the producer joinable, which is `std::terminate`.

**Your results are intact.** Plain `--vad` is still byte-identical pipelined vs serial (608 words), still `26/26 slices` pipelined, still 26 incremental progress lines. 1642 unit tests pass; CI green.

Two takeaways worth passing on, since neither is obvious:

- **Byte-identity on defaults proves the happy path only.** Your check was honest and I reproduced it — but the flags a split path silently drops are precisely the ones absent from a default invocation, so an equivalence claim needs the flag matrix.
- **"The two threads touch disjoint state" needs to cover what the caller does *between* the calls**, not just the two functions. Your audit of `encode_slice`/`decode_slice` was correct; the crash came from `finish_slice`, which the PR never touched.

Thanks again — the throughput win is real and the structure it landed in is good.

PR #357, opened 2026-08-14
fix(#356): split covering segment at gap-fill recoveries so output is monotone

### LINKED ISSUE #356: parakeet-ja: #89 gap-fill recoveries are emitted inside the segment they were recovered from — SRT output has backward-jumping, overlapping cues (0.8.28)
**Version:** 0.8.28 (git 99e990ae, Release, backends: cpu), Linux
**Backend:** parakeet, `parakeet-tdt-0.6b-ja-q8_0.gguf`, `-l ja`
**Reported downstream as:** SubtitleEdit/subtitleedit#13548 (full console log and generated subtitle attached there)

## Symptom

On a 633.6 s Japanese file, 18 of the 289 emitted transcript lines start **before the previous line ends**. These are not rounding errors — they are backward jumps of up to 10.5 s, and each one lands entirely inside a span an earlier segment already covers:

```
[00:00:32.660 --> 00:00:39.460]  まずねただちょっと私似
[00:00:39.460 --> 00:00:42.020]  てるところがあって1人で
[00:00:42.020 --> 00:00:42.980]  暮らすのは好き
[00:00:42.980 --> 00:00:43.300]  なんです。
[00:00:33.180 --> 00:00:34.540]  日本にもそんないない     <-- 10.1 s back
[00:00:34.540 --> 00:00:35.180]  です。
[00:00:36.860 --> 00:00:37.820]  えっと。
```

Same shape at ~68 s, ~254 s, ~281 s, ~367 s, ~379 s, ~426 s, ~430 s, ~442 s, ~461 s, ~486 s, ~513 s, ~544 s, ~564 s.

Command:

```
crispasr --backend parakeet -l ja -m parakeet-tdt-0.6b-ja-q8_0.gguf \
  -am wav2vec2-large-xlsr-53-japanese-q4_k.gguf \
  -f in.wav --output-srt --print-progress --max-len 35 --split-on-punct
```

VAD was auto-enabled by the parakeet backend (633.6 s > the 12 s JA window), producing 238 VAD segments → 71 slices.

## Root cause

The gap-fill second pass from #89 recovers text correctly but never reconciles it with the segment it was recovered *from*.

Reconstructing the slicing from the log: 238 VAD segments → offline post-merge → 42 → `crispasr_rechunk_slices` at the 12 s `vad_slice_cap_seconds` with ±2 s energy-minima cuts → 71. The slice covering the excerpt above is roughly **32.5 → 43.3 s**, and the first-pass parakeet call returns **one** segment spanning 32.66 → 43.30 whose words are sparse in the interior — exactly the #89 failure mode (the encoder blanks an utterance when enough context follows it). The give-away is the first split piece: 6.8 s of span for 11 characters, because `--max-len` split across the word-timestamp hole.

`crispasr_gap_fill_slice` (`examples/cli/crispasr_gap_fill.h`) then finds that hole, re-transcribes 33.18–35.18 and 36.86–37.82 in isolation, recovers them — and appends them as new segments:

```cpp
segs.push_back(std::move(rec));
recovered_any = true;
...
std::sort(segs.begin(), segs.end(),
          [](const crispasr_segment& a, const crispasr_segment& b) { return a.t0 < b.t0; });
```

The sort is on `t0` only. Nothing trims the enclosing segment's `t1`, and nothing splits it at the recovered span — so the enclosing segment keeps spanning 32.66 → 43.30 while its own recoveries sit inside it. Sorting by `t0` puts the enclosing segment first, and every recovery after it reads as a backward jump.

Nothing downstream catches this:

- `merge_segments` (`examples/cli/crispasr_run.cpp`) concatenates per-slice results in slice order with no sort and no monotonicity check.
- The `core_seg_hygiene` pass in the same function is env-gated (off by default) and, per its own comment, "only removes segments and rewrites text/t1 — it never reorders".

This also explains why `--force-aligner` made no difference for the reporter: the per-slice aligner at `crispasr_run.cpp` runs over `sl.start .. sl.end` and reassigns `seg.t0/t1` from the aligned words, so the enclosing segment still spans the whole slice.

Worth noting the recovered text is *correct* — #89's recovery is working. It is only unusable in `--output-srt` / `--output-vtt` because it is emitted inside a span that is already claimed.

## Suggested fix

In `crispasr_gap_fill_slice`, reconcile spans rather than just order. When a recovery lands inside an existing segment, split that segment at the recovered range (moving its words either side) instead of leaving it spanning the hole. Failing that, at minimum clamp each segment's `t1` to the next segment's `t0` after the sort.

A monotonicity assertion in `merge_segments` — unconditional, not behind the `core_seg_hygiene` env gate — would also have caught this at the source instead of it surfacing as a subtitle-rendering complaint.

## Downstream

Subtitle Edit now sorts and de-overlaps transcription results before its own post-processing (SubtitleEdit/subtitleedit#13626), so users will not see overlapping cues in the meantime. With that guard the affected lines come out *short* rather than overlapping — the underlying double-coverage is still there.

niksedk:
Fixes #356.

## What was wrong

The #89 gap-fill second pass recovers speech the parakeet-ja encoder blanks inside a slice, but its final step was:

```cpp
segs.push_back(std::move(rec));
...
std::sort(segs.begin(), segs.end(),
          [](const crispasr_segment& a, const crispasr_segment& b) { return a.t0 < b.t0; });
```

The first pass emits **one** segment whose sparse words straddle the hole, so its `t0..t1` encloses every recovery — a sort cannot fix that. The covering segment sorts first and every recovery after it reads as a jump backwards in time; `--output-srt`/`--output-vtt` carried cues overlapping by up to the hole length (18 of 289 segments on the reporter's 634 s Japanese file, worst 10.5 s — see the trace in #356).

## The fix

`crispasr_gap_fill_resolve_overlaps` replaces the plain sort:

- A covering segment is **split at each recovery**: words/tokens starting at or after the recovery's end move into a new tail segment; both halves rebuild their display text from their words (the same convention as the overlap-save trim in `crispasr_run.cpp`, CJK-space rule included).
- A covering segment with **no words past the recovery** is clamped at the recovery's start (handles the segment-level-only fallback and boundary-tolerance stragglers).
- Runs to a fixpoint, so a span enclosing several recoveries is split at each one. Every split strictly reduces the word count of the segment being split, so it terminates.

No text is added or lost; the list comes out in reading order: head words, recovery, tail words. The server path (`crispasr_server.cpp`) picks the fix up automatically since it calls `crispasr_gap_fill_slice`.

## Tests

`tests/test-issue-356-gap-fill-overlap.cpp` — scripted fake backend, pure CPU, no model load, labeled `[issue-356]`:

- end-to-end through `crispasr_gap_fill_slice`, mirroring the reporter's 32.66→43.30 s slice (fails on the previous sort-only code with exactly the reported symptom: a recovery starting inside the covering segment's span)
- span enclosing two recoveries → split at each, 5 monotone segments
- recovery before the covering segment's first word → covering segment becomes the tail
- word-less covering segment → clamped
- already-monotone input → unchanged
- no text lost when splitting

```
All tests passed (36 assertions in 6 test cases)
```

`test-issue-114-chunk-context-gate` (1157 assertions) and `test-issue-89-long-audio-fallback` (16 assertions) still pass; `crispasr-cli` (which compiles both header consumers) builds warning-free.

### DIFF
--- a/examples/cli/crispasr_gap_fill.h
+++ b/examples/cli/crispasr_gap_fill.h
@@ -14,6 +14,15 @@
 // Gate at the call site on backend.vad_slice_cap_seconds() > 0;
 // CRISPASR_GAP_FILL=0 disables, CRISPASR_GAP_FILL_MIN_CS tunes the trigger
 // (60 measured worse than the default 100: more variant noise, slower).
+//
+// Issue #356: a recovery lands *inside* the span of the segment it was
+// recovered from — the first pass emitted one segment whose sparse words
+// straddle the hole, so its t0..t1 covers the recovered range. Appending the
+// recovery and sorting by t0 alone leaves the covering segment first and every
+// recovery after it reading as a jump backwards in time (SRT/VTT cues overlap
+// by up to the hole length, 10+ s on the reporter's file). After each round
+// the covering segment is now split at the recovered range so the output is
+// monotone: head words, recovery, tail words.
 
 #pragma once
 
@@ -48,6 +57,66 @@ inline std::string crispasr_rebuild_text_from_words(const std::vector<crispasr_w
     if (!rebuilt.empty() && rebuilt[0] == ' ')
         rebuilt = rebuilt.substr(1);
     return rebuilt;
+}
+
+// Issue #356: make a gap-filled segment list monotone. The first pass emits a
+// segment whose sparse words straddle the recovered hole, so its span encloses
+// the recoveries; sorting by t0 alone leaves that segment first and the
+// recoveries overlap it. Resolve by splitting the covering segment at each
+// recovery: words (and tokens) that start at or after the recovery's end move
+// into a new tail segment, the head keeps the rest, and both rebuild their
+// display text from their words — the same convention as the overlap-save trim
+// in crispasr_run.cpp. A covering segment with no words on the far side of the
+// recovery is simply clamped at the recovery's start. Runs to a fixpoint so a
+// span enclosing several recoveries is split at each one; every split strictly
+// reduces the word count of the segment being split, so it terminates.
+inline void crispasr_gap_fill_resolve_overlaps(std::vector<crispasr_segment>& segs) {
+    const auto by_time = [](const crispasr_segment& a, const crispasr_segment& b) {
+        return a.t0 != b.t0 ? a.t0 < b.t0 : a.t1 < b.t1;
+    };
+    bool changed = true;
+    while (changed) {
+        changed = false;
+        std::sort(segs.begin(), segs.end(), by_time);
+        for (size_t i = 1; i < segs.size(); i++) {
+            crispasr_segment& prev = segs[i - 1];
+            crispasr_segment& cur = segs[i];
+            if (cur.t0 >= prev.t1)
+                continue;
+            // prev spans past cur's start. Split when prev still has words on
+            // the far side of cur; otherwise just cap prev's span.
+            if (!prev.words.empty() && prev.words.back().t0 >= cur.t1) {
+                crispasr_segment tail = prev; // copy metadata (speaker, chunk_id, ...)
+                std::vector<crispasr_word> head_words, tail_words;
+                for (auto& w : prev.words)
+                    (w.t0 >= cur.t1 ? tail_words : head_words).push_back(std::move(w));
+                std::vector<crispasr_token> head_tokens, tail_tokens;
+                for (auto& t : prev.tokens)
+                    (t.t0 >= cur.t1 ? tail_tokens : head_tokens).push_back(std::move(t));
+                tail.words = std::move(tail_words);
+                tail.tokens = std::move(tail_tokens);
+                tail.text = crispasr_rebuild_text_from_words(tail.words);
+                tail.t0 = tail.words.front().t0;
+                tail.t1 = tail.words.back().t1;
+                if (head_words.empty()) {
+                    // Nothing left of the recovery — prev *is* the tail.
+                    prev = std::move(tail);
+                } else {
+                    prev.words = std::move(head_words);
+                    prev.tokens = std::move(head_tokens);
+                    prev.text = crispasr_rebuild_text_from_words(prev.words);
+                    prev.t0 = prev.words.front().t0;
+                    prev.t1 = prev.words.back().t1;
+                    segs.push_back(std::move(tail));
+                }
+                changed = true;
+                break; // re-sort and re-scan
+            }
+            // No words past the recovery (or no words at all): clamp. A head
+            // word ending inside cur (boundary tolerance) resolves here too.
+            prev.t1 = std::max(prev.t0, cur.t0);
+        }
+    }
 }
 
 inline void crispasr_gap_fill_slice(CrispasrBackend& be, const whisper_params& params, const float* samples,
@@ -134,7 +203,9 @@ inline void crispasr_gap_fill_slice(CrispasrBackend& be, const whisper_params& p
         }
         if (!recovered_any)
             return;
-        std::sort(segs.begin(), segs.end(),
-                  [](const crispasr_segment& a, const crispasr_segment& b) { return a.t0 < b.t0; });
+        // Issue #356: a plain t0 sort left each recovery inside the span of
+        // the segment it was recovered from — split/clamp instead so the
+        // list comes out monotone.
+        crispasr_gap_fill_resolve_overlaps(segs);
     }
 }
--- a/examples/cli/crispasr_gap_fill.h
+++ b/examples/cli/crispasr_gap_fill.h
@@ -185,10 +185,25 @@ inline void crispasr_gap_fill_slice(CrispasrBackend& be, const whisper_params& p
                 return false;
             };
             for (auto& fseg : fill) {
+                const size_t first_kept = rec.words.size();
                 for (auto& w : fseg.words) {
                     const int64_t mid = (w.t0 + w.t1) / 2;
                     if (mid >= g.first - kCoverSlopCs && mid < g.second + kCoverSlopCs && !mid_is_covered(mid))
                         rec.words.push_back(std::move(w));
+                }
+                // The fill's tokens ride along with the kept words — the
+                // token-level outputs (JSON full, karaoke) read seg.tokens and
+                // a recovery with an empty list vanishes from them. A token
+                // joins at most one word, so a shared word boundary doesn't
+                // duplicate it; a token with no timestamp (t0 == -1) has no
+                // word to join and stays out.
+                for (auto& tk : fseg.tokens) {
+                    for (size_t i = first_kept; i < rec.words.size(); i++) {
+                        if (tk.t0 >= rec.words[i].t0 && tk.t0 <= rec.words[i].t1) {
+                            rec.tokens.push_back(std::move(tk));
+                            break;
+                        }
+                    }
                 }
             }
             if (rec.words.empty())
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2670,6 +2670,32 @@ catch_discover_tests(test-issue-114-chunk-context-gate
     PROPERTIES LABELS "unit;chunk-context;issue-114"
 )
 
+# ─── test-issue-356-gap-fill-overlap — issue #356 regression guard
+#
+# Pins crispasr_gap_fill_resolve_overlaps: the #89 gap-fill second pass used to
+# append recoveries inside the span of the segment they were recovered from and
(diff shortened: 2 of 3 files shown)


davideme:
Diagnosis matches what we hit on the non-JA side(https://github.com/CrispStrobe/CrispASR/commit/784c9c406397496e92f39e4e0f141d64c776d611).

**The sibling.** `gap_fill_segments` in `src/parakeet_orchestrate.cpp` (landed as 784c9c40, for #350) does the same find-holes / re-decode / merge-back job for the non-JA parakeet path. It deliberately did not reuse the #89 machinery, and the rationale is in the comment above the function, but the merge-back step is the part worth borrowing. It never appends a segment. Each recovered *word* is attached to the segment it belongs in (the last one starting at or before it, `parakeet_orchestrate.cpp:345`); that segment then re-sorts its words, rebuilds its text, and resets `t0`/`t1` from its own word list. Segment count and order are invariant, so the inversion #356 reports cannot arise, which is why this defect never surfaced on the non-JA path.

**Applied here**, that replaces the `rec` construction and `segs.push_back(std::move(rec))` at `crispasr_gap_fill.h:132` with a per-word placement plus a dirty-set rebuild after the round, and the trailing `std::sort` becomes unnecessary. Roughly 25 lines in place of `crispasr_gap_fill_resolve_overlaps`'s ~60, with no fixpoint loop to argue termination for, and no clamp fallback: the word-less covering segment case just doesn't exist when you never add a segment. You'd keep an append path only for a recovery landing before the first segment.

**The cost, stated plainly:** a covering segment ends up as one full cue spanning the hole instead of three. But that cue already spanned that range before the fix (that's the premise of #356), it was just sparsely populated. So insertion changes cue *duration* not at all versus today's output and only makes the cue complete, whereas splitting is a genuine improvement on it. If the shorter cues are the actual goal, `--max-len` already produces them from the word timings for every segment, not just gap-filled ones, and after insertion the recovered words are in the word list where that splitter can see them.

One thing the split has to decide that insertion never does: `crispasr_segment tail = prev;` copies the covering segment's metadata wholesale, so a diarized run gets one speaker turn duplicated across two cues. Probably harmless, but it's a decision with no obviously right answer.

examples/cli/crispasr_gap_fill.h
```diff
@@ -134,7 +203,9 @@ inline void crispasr_gap_fill_slice(CrispasrBackend& be, const whisper_params& p
         }
         if (!recovered_any)
             return;
-        std::sort(segs.begin(), segs.end(),
-                  [](const crispasr_segment& a, const crispasr_segment& b) { return a.t0 < b.t0; });
```
davideme, line 138:
```suggestion
        bool recovered_any = false;
        bool appended = false;
        std::vector<bool> dirty(segs.size(), false);
        for (auto& g : gaps) {
            // ... window setup and mid_is_covered unchanged ...
            auto fill = be.transcribe(samples + s0, s1 - s0, win0_cs, params);
            for (auto& fseg : fill) {
                for (auto& w : fseg.words) {
                    const int64_t mid = (w.t0 + w.t1) / 2;
                    if (mid < g.first - kCoverSlopCs || mid >= g.second + kCoverSlopCs || mid_is_covered(mid))
                        continue;
                    // Host = the segment with the greatest t0 at or before this
                    // word. Not "the last index at or before": a segment
                    // appended below sits at the end of the vector with a t0
                    // that may be earlier than an existing one.
                    size_t si = 0;
                    bool placed = false;
                    for (size_t i = 0; i < segs.size(); i++)
                        if (segs[i].t0 <= w.t0 && (!placed || segs[i].t0 >= segs[si].t0)) {
                            si = i;
                            placed = true;
                        }
                    // A word-less host carries text this rebuild would destroy,
                    // and no gap can land inside its span anyway (its whole
                    // span counts as covered above), so give the word its own
                    // segment instead.
                    if (placed && segs[si].words.empty() && !segs[si].text.empty())
                        placed = false;
                    if (!placed) {
                        segs.push_back(crispasr_segment());
                        dirty.push_back(false);
                        si = segs.size() - 1;
                        segs[si].t0 = w.t0;
                        appended = true;
                    }
                    for (const auto& tk : fseg.tokens)
                        if (tk.t0 >= w.t0 && tk.t0 <= w.t1)
                            segs[si].tokens.push_back(tk);
                    segs[si].words.push_back(std::move(w));
                    dirty[si] = true;
                    recovered_any = true;
                }
            }
        }
        if (!recovered_any)
            return;
        // Rebuild only what a word landed in. A segment nothing landed in keeps
        // the decoder's own detokenized text byte for byte.
        for (size_t i = 0; i < segs.size(); i++) {
            if (!dirty[i] || segs[i].words.empty())
                continue;
            crispasr_segment& s = segs[i];
            std::stable_sort(s.words.begin(), s.words.end(),
                             [](const crispasr_word& a, const crispasr_word& b) { return a.t0 < b.t0; });
            std::stable_sort(s.tokens.begin(), s.tokens.end(),
                             [](const crispasr_token& a, const crispasr_token& b) { return a.t0 < b.t0; });
            s.text = crispasr_rebuild_text_from_words(s.words);
            s.t0 = s.words.front().t0;
            s.t1 = s.words.back().t1;
        }
        if (appended)
            std::sort(segs.begin(), segs.end(),
                      [](const crispasr_segment& a, const crispasr_segment& b) { return a.t0 < b.t0; });
```

niksedk: Thanks for the close read! I'm keeping the split approach rather than merging recoveries into their covering segment, for two reasons:

- Merging yields one cue spanning the whole hole (10.5 s worst case on the reporter's file) with head + recovery + tail glued together; the split gives three correctly-timed cues, which is the better shape for the SRT/VTT output this issue is about.
- The byte-for-byte property already holds here: only segments that actually get split have their text rebuilt — untouched segments keep the decoder's text unchanged, same as in your sketch (and the merged host would be rebuilt from words either way).

The token observation is a real catch, though — recoveries have shipped with an empty `tokens` list since the #89 second pass landed, so they vanished from the token-level outputs (JSON full, karaoke). Fixed in b3c77340: each kept word now carries the fill's tokens in its span, and the #356 split already partitions tokens by time so split recoveries keep the right halves. Test extended to cover it.

CrispStrobe:
Merged as `d7d41bb7` — thanks. Verified live before merging on `parakeet-tdt-0.6b-ja-q8_0` with a 240 s Japanese clip: 4 cues starting before the previous one (worst **11.36 s** backward) → **0**, with the character multiset unchanged, so the split moves text without losing any. The regression guard is a real one: rebuilt against the pre-fix header, 5 of its 6 cases fail.

Three follow-ups landed on top, for the record:

**`c9f9356f` — clang-format.** The test came in with `  public:` / `  private:`; this repo's `.clang-format` sets `AccessModifierOffset: -4`, so `./tools/format.sh` (clang-format 18, what the Lint job runs) reported 2 violations. Worth knowing that a PR from a fork arrives with **zero CI signal** — every workflow on both of your commits sat at `action_required` and never ran, so the checks page being empty is not the same as green. Running `./tools/format.sh` and the affected test binaries locally is the way to catch that.

**`308cddfb` — the clamp fallback.** The split is right; the case it can't cover is when the covering segment's last word starts *before* the recovery ends, so `prev.words.back().t0 >= cur.t1` is false and there's nothing to move. Clamping `prev.t1` there moves no words, which leaves the word stream still non-monotone *and* strands a word's text in a cue that has already ended.

That matters because `crispasr_make_disp_segments` builds every SRT/VTT cue from **word** timestamps whenever a segment has words — `seg.t0`/`seg.t1` are unused on that path. So the clamp made the segment spans monotone while the actual subtitle output stayed backwards, and `require_monotone`, being a check on spans, passed it. It's reachable because the refill window runs to `g.second + kEdgePadCs`, so a recovered word can legitimately end ~0.2 s after the next first-pass word starts.

Those two now get merged and re-sorted rather than clamped — the same resolution `parakeet_orchestrate.cpp`'s `gap_fill_segments` and the session's `transcribe_ja_sliced` already use, which is why neither of them could produce #356 in the first place. It fires on real audio at `CRISPASR_GAP_FILL_MIN_CS=30`: 175 → 163 cues, identical character multiset, worst cue overlap 1.00 s → 0.20 s.

**`4edb2e10` + `9f1bddbb` — the guard the issue also asked for**, at `merge_segments`, the server loop, and `apply_session_hygiene` for the session ABI. It reports, never reorders.

Two small things in the split path also got fixed: `crispasr_segment tail = prev` copied `speaker_turn_next` onto **both** halves (the flag means "a turn follows this segment", so only the tail may keep it), and untimed tokens (`t0 == -1`) were discarded when the head came out empty.

Good PR — the diagnosis and the fixpoint/termination argument were both sound, and the guard was written to fail first. Thanks for it.

PR #361, opened 2026-08-15
chat: make the chat ABI cancellable, countable, and safe to tear down

mculbert:
### What this is

Twelve commits on the `crispasr_chat_*` C ABI, one per topic, each building and
passing on its own. Three paths aborted the host process rather than returning an
error; there was no way to stop a generation, no way to count tokens before decoding
one, and no way to close a session from another thread without freeing the model out
from under a running `llama_decode`. Shared-prefix KV reuse was written but never
reached. This fixes all of that.

Nothing here changes ASR. Every commit is scoped to the chat surface, except one
`/v1/chat/completions` fix in the bundled server and one `scripts/` chore.

### The three process aborts

Each killed the caller's process rather than returning an error, and each is now a
test.

**A prompt over 512 tokens.** `generate_loop` built one `llama_batch` over the whole
prompt. `llama_decode` does not error above `n_batch` — `llama-context.cpp:1583` is
`GGML_ASSERT(n_tokens_all <= cparams.n_batch)` and aborts. With the ABI's 512-token
default, any caller whose prompt ran past 512 tokens lost the process, and opening a
larger context did not help: `cparams.n_batch` is `min(n_ctx, params.n_batch)`. Now
walked in `llama_n_batch`-sized pieces, which is what the existing comment already
claimed.

**`crispasr_chat_memory_estimate`, on every model.** It loaded with `use_mmap` and
`no_alloc` both set; llama.cpp's mmap-backed buffer path opens with
`GGML_ASSERT(!ml.no_alloc)` (`llama-model.cpp:8619`). Nothing in the tree called it,
which is why it shipped. Fixed by dropping `use_mmap`, the pair llama.cpp's own
device-memory probe uses.

Making it reachable exposed two more defects in the same function, both fixed here:

- **The weights term was zero for any model over 2 GiB on Windows.** It read the file
  size with `fseek`/`ftell`, and `ftell` returns `long` — 32-bit under LLP64 — so a
  GGUF past 2 GiB could not be represented, `off > 0` was false, and the function
  reported success having omitted the model it was deciding whether to load. Now
  `std::filesystem::file_size`, and a size that cannot be read is an error rather than
  a zero.
- **The KV term used the requested `n_ctx`, not the one that gets allocated.**
  `llama_context`'s constructor does `GGML_PAD(cparams.n_ctx, 256)` before it
  allocates, so a request of 1000 tokens really takes 1024. The estimate now pads the
  same way.

**Teardown under load.** `crispasr_chat_close` took no lock at all — it freed the
context, the model and the session outright while another thread could be mid-decode.
See *Closing a busy session* below.

### What is new on the ABI

Two entry points and one error code.

- **`crispasr_chat_set_abort_callback`** — cancel an in-flight generation. The
  predicate returns `true` to continue, matching `whisper_encoder_begin_callback` on
  the ASR surface. `ggml-cpu` is the only backend exporting
  `ggml_backend_set_abort_callback`, so the loop also checks the hook itself, between
  prompt batches and before each sampled token; that is the only route on an
  accelerated decode. During *prefill* the scheduler splits a 512-token batch across
  the CPU backend too (183 graph splits at `bs=512` on Metal), so the in-graph check
  does fire there and cancels inside a batch rather than at its boundary.

  Cost: none worth reporting. 128 greedy tokens on `gemma-3-1b-it-Q4_K_M`, best of
  three, M4 Max — 0.541 s → 0.539 s GPU-offloaded, 0.777 s → 0.775 s CPU-only.
  Through ctypes on the CPU path the hook fired **581 times per token** for +0.1 %.

- **`crispasr_chat_count_tokens`** — the prompt size for a message array, through the
  model's own tokenizer and chat template. Sizing a context window previously meant
  estimating tokens from words, which holds for English and breaks for compounding,
  agglutinative and unspaced scripts. An empty message array counts whatever the
  template renders for no messages, which is template-dependent and is sometimes
  nothing: `mistral-v7`, `mistral-v1`/`v3`, `orion` and `minicpm` all write only from
  inside their loop over the messages and have no `add_ass` opening of their own, so
  they count 0. The header says so, and a table-driven test walks eight template
  families through the same model to hold it there.

- **`CRISPASR_CHAT_ERR_ABORTED`** — so a caller cannot read a cancellation as a decode
  fault, or a NULL return from the one-shot path as an allocation failure.

### One behaviour change on a shipped entry point

`crispasr_chat_generate` with `n_messages == 0` now returns an error. It previously
templated to an empty string, tokenized to a bare BOS, and generated from it. The
change arrived as the compensating guard for `apply_chat_template` no longer
short-circuiting `n == 0` — which is exactly the case `count_tokens` has to render. We
think an empty conversation has nothing to answer and the model continuing from an
empty assistant prefix is not a useful reading of the request, but it is a semantic
change and yours to reject; it is one `if` to drop.

### Performance

Shared-prefix KV reuse now works. It never did for any BOS-adding model: the prompt
was tokenised with `add_special` taken from `history.empty()`, so the first prefill
carried a leading BOS and every later one did not, the two sequences differed at index
0, and both the divergent and the append-only branches fell through to a full
re-prefill. The bundled interactive CLI was re-prefilling the entire accumulated
conversation every turn.

Measured on gemma-3-1b-it-Q4_K_M (Metal, M4 Max), five runs each, sub-millisecond
spread:

| Shape | Before | After |
|---|---|---|
| 214-token shared block, 1225-token prompt | 221.1 ms | 185.5 ms (−16 %) |
| 2072 of 2082 tokens shared | 405.6 ms | 30.5 ms |

**One dependency worth knowing about**, because it is invisible in the diff. The
truncating branch calls `llama_memory_seq_rm` and treats a `false` return as "the cache
refused, fall back to a full clear". In the vendored llama that `false` can only come
from `llama_memory_recurrent::seq_rm`; `llama_kv_cache::seq_rm` returns `true`
unconditionally (`llama-kv-cache.cpp:287`). So the fallback does not cover a
sliding-window cache, where truncating back past an evicted window would leave the
re-decoded suffix attending over a hole.

It does not bite today: `llama_context_default_params()` sets `swa_full = true`
(`llama-context.cpp:2894`) and the chat ABI never overrides it, so the SWA cache is
allocated at full `n_ctx` and nothing is evicted. Confirmed at runtime
(`creating SWA KV cache, size = 4096 cells` for a 4096 context on gemma-3-1b, whose
`n_swa` is 512), and three probes against a fresh session over a ~2,700-token shared
block came back byte-identical. But if `swa_full` is ever exposed through
`crispasr_chat_open_params` to save KV memory, or an upstream sync flips the default,
this path degrades silently. `llama.h:711` documents the check that would catch it
(`llama_memory_seq_pos_min`). We did not add it, since it guards a configuration that
does not exist; say the word and it is four lines.

### Closing a busy session

`crispasr_chat_close` freed the context, the model and the session outright under no
lock, so a close on a second thread pulled all three out from under a generation still
decoding on the first — and a generation is the one call long enough for that to be
the normal case rather than a narrow race.

The session mutex cannot be the guard: a generation holds it for as long as it
decodes, so a close that merely took it would still free the session under a second
call already blocked behind that generation, which acquires the mutex the instant the
close releases it.

Teardown therefore gets its own short-lived lock and a count of the calls currently
inside the session. Every entry point takes a `session_hold` for its span; close
retires the handle, waits for the count to reach zero, then frees.

**What it does not do, and the header now leads with this rather than burying it:**
make closing safe without the caller's own ordering. The refusal is decided by a check
that lives in the memory being freed, so a call that has read its handle but not yet
reached the lifetime lock can be freed underneath it and wake on a destroyed mutex,
and a second close blocked on that lock loses the same race. Closing must be ordered
against every other call on the same handle, exactly as for `fclose` and `llama_free`.
What close adds on top is the one thing a caller cannot do for itself: waiting for a
generation already inside.

A stronger version is possible — a process-global live-handle table with the liveness
check and the refcount bump under one lock, and an opaque token instead of a pointer so
a reallocation cannot alias — which would turn misuse into a defined refusal rather
than undefined behaviour. It is not here because it changes what the opaque handle *is*,
which is a larger ABI commitment than anything else on this branch, and because nothing
else in CrispASR's C surface (or in the vendored llama.cpp and whisper.cpp beside it)
works that way. Yours to want.

### The server

`/v1/chat/completions` runs two C calls per request on one process-wide session —
a reset, then a generate. The session's own mutex serialises each call but not the
pair, so two overlapping requests could interleave, with one request's reset landing
between another's reset and its generate; the second then prefills onto a cache it did
not clear and answers with the first request's conversation in context. One mutex now
spans the whole request. This costs no throughput that was real, since the generate
already held the session mutex for its whole decode.

### `scripts/sync-llama.sh`

Now records the upstream revision it copied from, mirroring `sync-ggml.last`. The seed
value was derived by comparing all 169 vendored files against upstream by blob hash; it
resolves to a 19-commit window and `6b10a82c` is recorded as the earliest.

### Testing

Every gated suite self-skips without a model. Counts with `CRISPASR_CHAT_TEST_MODEL`
set to `gemma-3-1b-it-Q4_K_M`, on an M4 Max:

| Suite | Cases |
|---|---|
| `test-chat-ggml` | 424 assertions in 11 cases |
| `test-chat-abort` | 64 assertions in 7 cases |
| `test-chat-close` | 24 assertions in 3 cases |

Also clean: `clang-format` at the CI-pinned 18, `cppcheck` with the lint job's flags,
and a build of `crispasr-lib` and `crispasr-cli`.

Two test-design notes, because both cases passed for the wrong reason first:

**The prefill-abort case needed a scaling oracle, not a count.** Allowing one
consultation and asserting the run stopped after more than one passes even with the
between-pieces check deleted, because the in-graph hook aborts inside batch one
instead. What separates them is how the count scales: prefilling one prompt in three
pieces costs three seam checks plus three graphs' worth of in-graph ones, because a
graph's node count is a property of the model and not of the batch. Measured: 6 and 2
with the check, 4 and 2 without.

**The prefix-reuse cases compare against a fresh session at temperature 0 and require
byte-identical output.** That is the only oracle that catches a cache holding something
the history does not describe.

`test-chat-close` parks a generation inside its own token callback so the overlap is
deterministic with no sleeps. Removing the wait from `close` segfaults it. Two things
it deliberately does *not* assert are recorded in the file: that the generation
returned before the close did — `in_use` falls in a destructor as the C function
unwinds, before it returns to its caller, so close is free to return first, and
asserting otherwise would flake — and anything about a thread paused before it reaches
the lifetime lock, which cannot be tested from inside the memory being freed.

---

### 1. The streaming path contradicts the header under `stop`, and I did not fix it

`crispasr_chat.h` says output is truncated BEFORE the match. The one-shot path does
that; the streaming path does not. `out.append(piece)` and `on_token(piece)` both run
before `find_first_stop` scans and truncates, so the callback has already received the
piece the match lands in — the streamed concatenation is the one-shot text plus that
chunk. Measured: one-shot `"1\n2\n3\n"` against streamed `"1\n2\n3\n4"`.

I left C alone because the obvious fix is incomplete. Moving the scan above `on_token`
handles a match **within** one piece but not one that **straddles** pieces, which the
comment claims to honour: with stop `"45"` and pieces `"4"`, `"5"`, the `"4"` is
already delivered when `"45"` matches. A complete fix has to withhold trailing bytes
that could still become a stop prefix — buffering up to (longest stop − 1) — which adds
delivery latency and changes behaviour for every existing streaming consumer, including
the server's SSE path. That seemed yours to decide.

### 2. Should `n_messages == 0` be an error?

See *One behaviour change* above. Could revert and let an empty conversation
generate from a bare BOS as before.

### 3. Is `swa_full` ever going to be exposed?

See *Performance* above. If the answer is no, the prefix-reuse path is sound as
written. If it might be, the `llama_memory_seq_pos_min` guard should go in while
the reasoning is fresh.

---

## Found while working, not fixed here

Outside this branch's scope, recorded rather than silently left. None is caused by
this branch, and none blocks it.

**`cargo test -p crispasr` cannot run unaided.** `crispasr-sys/build.rs` emits its
rpath via `cargo:rustc-link-arg`, which cargo does not propagate to a dependent's
binaries, so the test executable links with zero `LC_RPATH` entries and dies with
`Library not loaded: @rpath/...`. `cargo:rustc-link-arg-tests` / `-bins` would fix it.

**`release-wrappers.yml:141` claims a CI check that does not exist.** The comment says
`cargo check` on both crates runs "on every push", and it is load-bearing — it is the
stated reason publishing can use `--no-verify`. There is no `cargo` invocation in any
workflow in `.github/workflows/`.

**`bindings-go.yml` does not trigger on the chat ABI.** Its path filters cover
`bindings/go/**` and `crispasr.h` but not `include/crispasr_chat.h` or `src/chat.cpp`,
so a change to this ABI can merge without the Go binding compiling. Two lines, but it
is your file, so it is a suggestion rather than a commit.

**A multi-element `LABELS` in `catch_discover_tests` destroys `SKIP_RETURN_CODE`.** The
`PROPERTIES` block is spliced in unquoted and Catch2 appends its own skip code, so a
two-element label list desynchronises the pairing and discards it — a gated test then
reports **Failed** instead of Skipped. **184 registrations** in `tests/CMakeLists.txt`
have this shape. Verified: with `LABELS "unit;live"` the generated file reads
`LABELS unit live SKIP_RETURN_CODE 4`, `ctest -L live` finds zero tests, and a
model-less gated case goes red. An escaped `"unit\;live"` does not survive either.

**`clang-format` runs over vendored third-party code.** `b9081285` reformatted
`examples/talk-llama/` with this repo's `PointerAlignment`, the opposite of upstream
llama.cpp's. Cosmetic, but it destroys byte identity with upstream, so the vendored
revision can no longer be identified by comparing blobs, and the next `sync-llama.sh`
run will produce a whole-tree diff. Excluding that directory from the format gate would
preserve it.

**The vendored ggml and the vendored llama.cpp are about five months apart.** ggml is
0.17.0 from `404fcb9d`; the llama.cpp sources are from 2026-03-15/16. That skew already
broke the build once — ggml v0.17 added a trailing `int64_t K` to
`ggml_gated_delta_net` and `models/delta-net-base.cpp` had to be forward-ported by hand
— and it was caught only by building the full CLI, not by the ggml-only CI job.

**Two pre-existing Rust test failures**, verified against a tree without this branch's
work: `kokoro_lang_helpers` and `vad_slices_null_model`.

**Pre-existing failures elsewhere**: `gofmt -l bindings/go/` reports `whisper.go` and
`examples/go-model-download/context.go`;
`test_backend_config.py::test_detect_backend_from_gguf_returns_name` expects
`vibevoice` and gets `vibevoice-tts`; repo-wide `dart format --set-exit-if-changed` is
red on `lib/src/crispasr.dart` and `test/beats_live_test.dart` under the 3.13 formatter.

Authored by Claude Opus 5
Adversarial review by GPT 5.6 Sol


CrispStrobe:
Merged as `f9de92b2`. This is the most carefully written PR the repo has had — three process aborts, each root-caused to a specific upstream assert with a file and line, each turned into a test, plus a section on what the fix deliberately does *not* guarantee. The two "passed for the wrong reason first" notes are the part I'd most like other contributors to copy.

**Verified independently rather than from the description:**

- All three gated suites pass here on a **different model** than yours — `smollm2-360m-instruct-q8_0` rather than gemma-3-1b — with assertion counts matching yours exactly: 24/3, 64/7, 424/11.
- **Red-verified the headline fix.** Deleting `s->idle.wait(lk, ...)` from `crispasr_chat_close` and rebuilding makes `test-chat-close` die (SIGABRT here rather than your segfault — same defect, different landing). The guard is real, not decorative.
- Audited the teardown design against the code rather than the prose: `life` + `retiring` + `in_use` + condvar is sound within the contract you state, `close` frees outside the lock, and a hold taken during retirement is refused. The residual window — a call descheduled after reading its handle but before locking `life` — is real and is exactly what the header now leads with.
- The trap I went looking for, `set_abort_callback` taking `s->mu` while a generation holds it for its whole decode, is already spelled out in the header ("Register before starting a generation and have the callback read your own flag"). Good.

**Your three design questions:**

1. **Streaming vs one-shot under `stop`.** Agreed, and agreed it is not a one-liner. Withholding (longest stop − 1) bytes is the only complete fix and it changes latency for every existing SSE consumer. Leaving C alone and documenting was the right call; worth a separate issue rather than a quiet fix.
2. **`n_messages == 0`.** Keep it. I checked the only in-tree caller: `/v1/chat/completions` already rejects an empty `messages` array with a 400 (`missing or empty 'messages' array`) before it ever reaches the ABI, so nothing user-visible changes and `count_tokens`'s contract stays coherent.
3. **`swa_full`.** Leave the guard out. Your reasoning holds — it guards a configuration that does not exist — and the note in the PR is what makes it recoverable if `swa_full` is ever exposed. If it is, that's the commit that should add the four lines.

**One thing to know:** this branched from `cc498701` and merged clean, but that base is now seven commits behind — `main` has since taken #353's parakeet pipelining and #356's segment time-order guard, both of which touch `crispasr_server.cpp`. The regions are far apart and the merge was clean, but your green CI was measured against the older base.

Follow-up landed in `462d8a6d`, mostly prompted by your own "found while working" section — see the #362 thread.

Thanks for this one.

Issue #143, opened 2026-06-03
Server Mode Does not Support Diarization


Matir:
As far as I can tell, the server mode does not support diarization. I've tried enabling it via a `--diarize` command-line flag, passing `diarize=true` in the request, ensuring I'm getting verbose JSON response, etc.

I spent some time looking at the source code and it doesn't seem that parameter gets passed along anywhere, so I'm not seeing anything I'm doing wrong.

Is there a way to support diarization in the server?

CrispStrobe:
Fixed in 59d22400 — server now accepts `diarize=true` and `diarize_method=energy|xcorr|vad-turns|pyannote|sherpa` on both `/inference` and `/v1/audio/transcriptions`. Also wired up all other previously missing CLI features (translate, punctuation, VAD, hotwords, grammar, whisper sampling params). See `docs/server.md` for the full field list.

PR #362, opened 2026-08-15
chat: bind the chat ABI from Python, Go, Java and Dart on one contract

mculbert:
### What this is

One commit per language: Python, Go and Java chat bindings that did not exist, and the
Dart one finished. (Rust is in #361, alongside the C, because `crispasr-sys` transcribes
the header and the two move together.)

Nothing here touches C, ASR, or any workflow.

### Common contract

- **The abort predicate returns `true` to continue**, matching the header.
  Each binding's test pins **both** directions, because either
  half alone passes under an inverted implementation.
- **A token-callback failure cancels the generation**, and the predicate is not
  consulted again.
- **An interior NUL is rejected naming the field**, not silently truncated where C stops
  reading — including in the model path.
- **A partially-filled params object keeps the ABI defaults.**
- **Only `CRISPASR_CHAT_ERR_ABORTED` is switched on**, with a fallback to the call's
  return value, per the header's rule that every other code is a diagnostic.
- **No exception unwinds through C** — captured, answered as abort, rethrown after the
  call.

All four assert the **same literals** for the same prompt and sampler settings — stop
`["5"]` truncating `1\n2\n3\n4\n5\n6\n7\n8\n` to `1\n2\n3\n4\n`, and stop `["7","4"]`
and `["4","7"]` both giving `1\n2\n3\n`, so the earliest match in the output wins rather
than the first array entry. Four independent marshallings of one C feature, agreeing
byte for byte. (Rust in #361 asserts the same set, making five.)

### Two bugs worth calling out, because their tests passed while they were live

**Streamed chunks were corrupted for non-Latin output** in Python. The C side hands the
callback the raw bytes of `llama_token_to_piece`, so a byte-fallback token is one byte
of a multi-byte character; decoding each fragment independently turned it into
replacement characters. Every binding documented and tested that concatenating the
chunks reproduces the one-shot reply — and every test passed, because every suite used
English prompts. Gemma 3 holds common CJK and emoji whole; the regression test needs
rare code points.

Go had the same shape with a different symptom, and a later review pass caught it: a Go
`string` carries arbitrary bytes, so concatenation was already exact and nothing was
corrupted — but each individual chunk the caller received was not valid UTF-8, and a
JSON or SSE consumer replaces or rejects those before they are ever joined. Measured,
streaming `🪿🫏🪼` on Gemma 3: twelve callbacks, twelve single bytes, every one invalid
on its own. Go now assembles like the others, and its test asserts `utf8.ValidString`
per chunk and that there are no more chunks than characters — the concatenation
equality alone passes either way, which is why it did.

**A Go params struct silently discarded the ABI defaults.** Every field of a caller's
struct was copied over the defaults, so setting one option zeroed the rest —
`&ChatGenerateParams{MaxTokens: 64}` passed temperature 0, greedy decoding in place of
0.8.

### Lifetime, and why each binding solves it differently

#361's `crispasr_chat_close` waits for calls already inside the session, but it cannot
see a call that has read its handle and not yet entered C. Supplying that ordering is
the caller's job — which here means each binding's. They are not five copies of one
workaround; each uses what its language already gives it:

- **Python** counts. It deliberately does not hold one lock across every call, so
  `count_tokens`, `reset` and the read-only accessors stay pollable during a
  generation; a short-lived guard counts who holds the handle and `close` waits for
  that count to drain. `generate`/`generate_stream` additionally take a non-blocking
  call lock and **raise rather than queue**, because two threads sharing one session is
  a caller error the ABI names and waiting does not turn into concurrency.
- **Go and Java** hold one lock per session across every call, `Close`/`close()`
  included.
- **Dart** needs nothing: FFI calls run synchronously on the isolate that made them, so
  a generation and a close cannot overlap, and a handle is not a transferable value.
  That is recorded at `close()` so the absence reads as a conclusion rather than an
  omission.

Two lifetime defects fixed along the way:

**Dart could free past its own stop array.** `_marshalParams` allocated from
`params.stop.length` and stored that count in the C struct; cleanup re-read the caller's
list. `shouldContinue` runs synchronously inside the native call and is forbidden only
from re-entering the same session, so that list is legally mutable mid-generation —
appending to it made cleanup walk two pointers past a two-slot array and free whatever
was there. Reproduced: abort trap 6. Marshalling now returns an owned object holding the
allocations and the frozen count, and cleanup never consults `ChatGenerateParams` again.

**Dart's free-on-GC finalizer re-opened the library by default name** rather than using
the session's own, so a session opened from an explicit path could be freed through a
different library than it was created by.

**Java passed model paths as JNA `String`**, which converts through the configured
native encoding — a legacy code page on the Java 8 Windows target this binding supports
— while the C side reads UTF-8. Both entry points now marshal the path with the
binding's own `ChatNative.cstring`, as they already did for everything else.

One related gap is recorded rather than fixed, because it is not this branch's: on
Windows, `crispasr_chat_memory_estimate` reads the path through
`std::fopen`/`std::filesystem::path`, which use the ACP, while `ggml_fopen` — how llama
actually opens the model — treats it as UTF-8. The two disagree for a non-ASCII path.
Fixing it means picking a convention for the whole C surface, which is yours.

### Bindings not included

Ruby, C# and JavaScript. We have no working toolchain to build and test them here, and
an untested binding is worse than an absent one. Each would be an additive commit on
the same pattern; nothing already written has to change.

### Testing

Every gated suite self-skips without a model. Against `gemma-3-1b-it-Q4_K_M` on an M4
Max:

| | Result |
|---|---|
| Python | 34 tests, all pass |
| Dart | 19 tests, all pass |
| Go | `go vet` clean; **suite not run here** — see below |
| Java | **not run here** — see below |

Two honest gaps, both environmental rather than in the code:

- **Go's test binary would not link in this checkout.** The cgo `LDFLAGS` name
  `../../build/...` and enumerate every conditionally-built sub-library; the build tree
  here was configured without the codec deps, so the link fails on `-lopus`. `go vet`
  compiles the package cleanly, and `gofmt` is clean, but the 26 chat cases were last
  run against a full build earlier in this branch's history rather than against this
  exact tree.
- **Java needs a JDK 20 or older.** Gradle 8.1 rejects class file major version 70
  (JDK 26), which is what is installed here. The 52 cases were last run against a
  staged build earlier in this branch's history. Note also that
  `bindings/java/build.gradle` stages libraries from a hard-coded `../../build/src`, so
  a checkout that builds anywhere else fails every model-backed case with
  `UnsatisfiedLinkError` — the `copy*` tasks silently copy nothing and `test` runs
  anyway.

If you would rather not take a binding whose suite has not been re-run against the exact
tree, drop the Go and Java commits and take Python and Dart; the four are independent
and each is one commit.

---

### 1. The Java session is `AutoCloseable` with no GC backstop

`ChatSession` frees on `close()` and nowhere else, matching the existing Java binding's
style — a caller who forgets try-with-resources holds the model and its KV cache until
the process exits. A `Cleaner` would make that survivable, but `java.lang.ref.Cleaner`
is Java 9, and this binding targets 8. I left it as it is because matching the
surrounding code seemed better than raising the baseline unilaterally.

### 2. None of these four has a CI job that can fail on them

Go's workflow exists but does not trigger on the chat ABI (#361 records the two path
filters that would fix it, deliberately not committed since it is your file). There is
no Python test job. The `bindings-java` job lives in `build.yml`, whose path filter
covers neither Java nor `src/chat.cpp`, and its Gradle step carries
`continue-on-error: true`. No PR job runs `dart analyze` or `dart test` over the Flutter
package.

This is pre-existing — `main` already ships a 3,772-line Python binding and a 3,318-line
Rust crate with no test job, and the Python binding has never had a test file.
It is a note that the byte-for-byte agreement this PR is built on is only durable if
something re-checks it. Whether to add or repair those jobs is yours; I have not touched
any workflow.

### 3. The streaming/stop divergence is pinned, not fixed

#361 question 1 explains why. All four bindings here inherit the unqualified promise
that streamed and one-shot text agree, all four tested it, and every test passed because
none used a stop sequence. That is now corrected and **both** strings are pinned in all
four. If you decide C should conform to the header instead, the new expectation is one
line per binding.

---

## Found while working, not fixed here

Not caused by this branch. The C-side list is in #361; these are binding-side.

**The ASR surface has the same callback-lifetime bug this PR avoids for chat.**
`Session._progress_cb_holder` and `Mic._cb_holder` each hold a ctypes trampoline in a
single instance attribute, which is its only strong reference. Two threads on one
`Session` frees the running call's progress trampoline while C holds the pointer. Same
remedy as the chat binding uses: build it per call and hold it in the calling frame.

**`Session.transcribe_chunked` clears its progress callback with a literal `None`**,
which CPython 3.14 rejects for a function-pointer argtype with `ArgumentError`.

**`_get_lib()` is undefined** in the three `watermark_*` functions in
`python/crispasr/_binding.py` — every call raises `NameError`.

**`python/crispasr/_binding.py` imports numpy at module scope**, so pytest fails at
*import* on an environment without it rather than skipping the tests that need it. This
one obstructs the chat binding's own testability: nothing in the chat suite needs numpy,
but nothing in the suite runs without it.

**`bindings/java/src/test/.../ChatSessionTest.java` reaches past the wrapper.**
`rawChunkBytes` calls `ChatLib.INSTANCE.crispasr_chat_open` directly to observe the exact
bytes C delivers, so the signature change for UTF-8 paths above had to be made there
too. Worth knowing if you move that declaration again.

**The `bindings-java` CI job cannot fail.** Its `Build` step carries
`continue-on-error: true` ("JNA struct alignment on Win64, #143"), so a genuine
regression is invisible. It also downloads only `tiny.en`, so nothing chat-related runs
in it.

**`NativeCallable.isolateLocal` does accept a closure under `dart test`.** The Dart
binding's recorded reason for not binding streaming was false; streaming is bindable and
simply not bound. Left as it is here, since binding it is a feature rather than a fix.

Authored by Claude Opus 5
Adversarial review by GPT 5.6 Sol


CrispStrobe:
Merged as `f409a5b2`. Five bindings on one contract, each with its own suite, and the design decisions that matter are right: params seed from `*_params_default` rather than zero-init, Go binds the real C struct through cgo so it cannot drift, and the Python ctypes mirror matches `crispasr_chat_generate_params` field-for-field — I checked all eleven fields against the header.

**Follow-up landed in `462d8a6d`**, and it starts from your own "found while working" list.

**CI compiled one of the five bindings.**

- `bindings-go.yml` fired for this PR because it adds `bindings/go/chat.go` — but its filters are `bindings/go/**` and `crispasr.h`, so a *later* change to `include/crispasr_chat.h` or `src/chat.cpp` would not have run it. That is the gap you reported; the two lines are in.
- `bindings/java` is built only by `build.yml`, whose PR filter is `ggml` / `.gitmodules`. The `android` check that ran here is `ci.yml`'s app build, not the binding — so 1,566 lines of Java binding and 914 of Java tests compiled in nothing.
- No PR-time `cargo` or `dart` job exists at all, which also confirms your `release-wrappers.yml` finding from the other thread.
- `tests/test_python_chat.py` — 891 lines — is named by no workflow, and `regression.yml` pytests two files *by name* rather than scanning a directory. It could only ever run by hand.

Rather than stand up five toolchains, the follow-up takes the cheap half: `tools/check-chat-binding-coverage.py` parses the header for its `CRISPASR_CHAT_API` functions and asserts each is named in each binding — the idiom `test-copies-in-sync.cpp` and `check-backend-wiring.py` already use here. **All five bind all 14 entry points**, so this PR is complete on that axis. Red-verified by adding a 15th to the header: all five report it with the file to fix.

It checks omission, not signatures — a binding with a wrong argument type still passes. That is what your per-language suites are for.

The Python suite now runs in the no-model job: with neither a library nor a model it still exercises the ctypes structs, the param defaults and the UTF-8 chunk assembly, and skips the rest.

**One portability issue worth your attention, since it affects four suites I did not touch.**

The stop-sequence literals are one model's greedy reply, but the gate accepts any small chat GGUF — `test_python_chat.py`'s own docstring names three, including `smollm2-360m`. On smollm2 that prompt answers `"1 2 3 4 "` with spaces rather than newlines, so four cases failed on the *separator* while every behavioural assertion beside them passed. A red for a reason unrelated to the code under test.

I did not delete the literals — the cross-binding byte-identity oracle is the most valuable thing about them, and it only works if all five pin the same string. Instead `TestChatStopSequences` now generates the baseline reply once in `setUpClass` and skips with an explanatory message when it is not the model those strings describe. 4 failed → 0 failed, 29 passed, 5 skipped; the no-model path is unchanged.

**The same literals are hardcoded in `chat_test.go`, `ChatSessionTest.java`, `integration.rs` and `chat_test.dart`** and have the same issue. I only fixed Python, because it is the one now wired into CI, and because the right shape in each language is yours to pick. The `ChatSessionTest.java` comment ("The reply this model gives is...") already half-acknowledges it.

Thanks — this and #361 together are a lot of careful work.

Issue #326, opened 2026-07-30
Question: Diarization seems excessively slow


logiclove:
Adding diarization change the performance from `70x` to `26.3x` realtime. Is this normal ?
* Logs
```
crispasr      | crispasr-server: transcribed 2888.2s audio in 109.93s (26.3x realtime), format=verbose_json
```

* Settings used:
```
CRISPASR_MODEL: ${CRISPASR_MODEL:-/models/cohere-transcribe.gguf}
CRISPASR_BACKEND: ${CRISPASR_BACKEND:-cohere}
CRISPASR_LANGUAGE: ${CRISPASR_LANGUAGE:-auto}
CRISPASR_EXTRA_ARGS: ${CRISPASR_EXTRA_ARGS:---vad --frequency-penalty 0.85 
  -t 12
  --lcs-dedup on --lcs-min-length 4 --split-on-punct  --lid-backend silero 
  --diarize --diarize-method pyannote --sherpa-segment-model auto 
  --diarize-embedder auto --diarize-cluster-threshold 0.5 --diarize-max-speakers 4 }
```

* Hardware
`lscpu`
```
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             48 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      24
  On-line CPU(s) list:       0-23
Vendor ID:                   AuthenticAMD
  Model name:                AMD Ryzen 9 9900X 12-Core Processor
    CPU family:              26
    Model:                   68
    Thread(s) per core:      2
    Core(s) per socket:      12
    Socket(s):               1
    Stepping:                0
    Frequency boost:         enabled
    CPU(s) scaling MHz:      32%
    CPU max MHz:             5662.0161
    CPU min MHz:             613.9540
    BogoMIPS:                8782.91
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good amd_lbr_v2 nopl xtopology nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy svm e
[...]

Issue #107, opened 2026-05-18
Native pyannote path uses segmentation-only labels as diarization, producing unstable speaker IDs


hcoona:
## Summary

The native `--diarize-method pyannote` path currently treats pyannote segmentation output as if it were complete speaker diarization. After fixing the runtime issues that caused crashes and obvious class collapse, the method can produce more than one speaker label, but the labels are still visibly incorrect on multi-speaker long-form audio.

The likely design issue is that pyannote segmentation provides local frame-level speaker activity classes, not globally consistent speaker identities. The current implementation maps those local classes directly onto ASR segments without speaker embeddings, clustering, cross-window label stitching, or splitting long ASR segments at speaker turns.

## Actual behavior

Running native pyannote diarization on a two-speaker podcast-like file produces output with multiple speaker labels, but the assignments are not reliable or semantically consistent:

```bash
../build/bin/crispasr \
  --backend cohere \
  -m auto \
  -f 3bnLgOxHMB8.wav \
  -l en \
  --vad \
  -am auto \
  --diarize \
  --diarize-method pyannote \
  --sherpa-segment-model models/pyannote-seg-3.0.gguf \
  -ovtt \
  -ojf \
  --split-on-punct \
  -of podcast_cohere_pyannote
```

After local runtime fixes, the same file produced a non-collapsed distribution:

```text
segments: 74
speakers: {'(speaker 1) ': 67, '(speaker 0) ': 7}
```

However, manual inspection shows the speaker assignments are still clearly wrong for the conversation. The issue is no longer simply "all segments are the same speaker"; the native pyannote path appears to lack the rest of the diarization pipeline needed to make speaker IDs meaningful and stable.

## Expected behavior

If the CLI presents `--diarize-method pyannote` as speaker diarization, the output speaker labels should represent stable speaker identities across the file, and speaker changes inside long ASR spans should be handled appropriately.

At minimum, if the native GGUF path is currently only a segmentation-based heuristic, the CLI and documentation should make that limitation explicit so users do not expect full pyannote-style diarization quality from a segmentation-only model.

## Reproduction steps

1. Build CrispASR with native pyannote GGUF support.
2. Use a long-form two-speaker audio file, such as a podcast interview.
3. Run:

```bash
../build/bin/crispasr \
  --backend cohere \
  -m auto \
  -f 3bnLgOxHMB8.wav \
  -l en \
  --vad \
  -am auto \
  --diarize \
  --diarize-method pyannote \
[...]

Issue #98, opened 2026-05-17
pyannote diarization segfaults on CUDA builds due to CPU dereference of GPU-resident GGUF tensors


hcoona:
## Summary

`--diarize-method pyannote` segfaults on CUDA builds because the native pyannote segmentation runtime loads GGUF weights onto the selected GPU backend, while `pyannote_seg_run()` still directly dereferences `ggml_tensor->data` from CPU-side manual F32 loops.

## Actual behavior

Running pyannote diarization on a CUDA-enabled build crashes with `SIGSEGV` during the first SincNet convolution weight read in `src/pyannote_seg.cpp`.

GDB points to:

```cpp
sum += sinc[co * K0 + k] * samples[t * S0 + k];
```

The faulting pointer is `m.sinc_filters->data`, and the pyannote model backend is `CUDA0`.

## Expected behavior

`--diarize-method pyannote` should either:

1. run successfully on CUDA-enabled builds by keeping the pyannote segmentation weights in CPU-accessible memory, or
2. explicitly support GPU execution by using ggml graph/backend APIs rather than direct host dereferences of backend tensor storage.

At minimum, a CUDA-enabled build should not crash when pyannote diarization is requested.

## Reproduction steps

From the repository root, with a CUDA-enabled build and a pyannote GGUF segmentation model available:

```bash
cd workspace

../build/bin/crispasr \
  --backend cohere \
  -m auto \
  -f test60.wav \
  -l en \
  --vad \
  -am auto \
  --diarize \
  --diarize-method pyannote \
  --sherpa-segment-model models/pyannote-seg-3.0.gguf \
  -ovtt \
  -ojf \
  -of test60_pyannote
```

Under GDB:

```bash
gdb -q --batch \
  -ex 'set pagination off' \
  -ex 'run' \
  -ex 'bt full' \
  --args ../build/bin/crispasr ...
```

Observed crash:

```text
Thread 1 "crispasr" received signal SIGSEGV, Segmentation fault.
pyannote_seg_run (...) at src/pyannote_seg.cpp:237
237    sum += sinc[co * K0 + k] * samples[t * S0 + k];
```

## Environment

```text
CrispASR version: 0.6.6
Build type: RelWithDebInfo
Backends: cpu,cuda
GPU: NVIDIA GeForce RTX 4090
CUDA backend selected: CUDA0
OS: Linux
Model: pyannote-seg-3.0.gguf
ASR backend used in repro: cohere
```

Relevant runtime log:

[...]

Issue #334, opened 2026-08-05
[v0.8.25] CosyVoice-TTS - long delay, pitch shifting and accent issues


subof:
Hi,
I am experiencing issues with voice cloning using the cosyvoice3-tts backend (model: cosyvoice3-0.5b-2512.gguf) integrated into Subtitle Edit. The output pitch and cloning quality change drastically depending on the reference audio sample rate, suggesting an issue with sample rate handling or resampling in the CrispASR backend.
Observed behavior:

   1. With 16.0 kHz Reference Audio:
   * Specs: PCM, 16.0 kHz, 16 bits, 1 channel, 256 kb/s, ~7s duration.
      * Result: The cloned voice has a very heavy, unnatural foreign accent, likely because the 16 kHz audio is fed into the 24 kHz model without proper resampling, corrupting phoneme alignment.
   2. With 24.0 kHz Reference Audio:
   * Specs: PCM, 24.0 kHz, 16 bits, 1 channel, 384 kb/s, ~20s duration.
      * Result: Pronunciation improves, but a severe "chipmunk effect" occurs. The output pitch is significantly shifted upward, which points to a mismatch in the output vocoder sample rate or time-step scaling inside the C++ container.
   
Expected behavior:
The backend should automatically resample input reference audio to the model's native sample rate and generate the output voice with the correct pitch and speed, regardless of the input file's parameters.
Could you please check how the sample rate and time-steps are managed for the CosyVoice3 backend? Thank you!

I tested it on Windows, HP Z8 G4 RTX 3090TI

subof:
p.s. The basic model Fun-CosyVoice3-0.5B-2512 is the standard pre-trained version, while the RL modification has undergone additional Reinforcement Learning (RL).
The RL version is designed specifically to achieve maximum speech quality, accuracy, and generation stability.

Could you also add the RL version to the backend for better voicing, please?
Thank you.

CrispStrobe:
Thanks — this reproduced, though the cause turned out not to be the sample rate. Four things are fixed on `main`, and the RL model is published.

[...]

PR #364, opened 2026-08-16
feat(diarize): onnxruntime backend for pyannote-seg/TitaNet (CUDA EP with CPU fallback)

tilllt:
## What

Adds an optional onnxruntime execution path for the two neural nets used by `--diarize-method pyannote`: the pyannote segmentation model and the TitaNet speaker embedder. Model paths ending in `.onnx` are executed by onnxruntime instead of the CPU-only ggml runtime. The CUDA execution provider is attached automatically when the linked onnxruntime distribution provides it (use `onnxruntime-gpu`), with a silent fallback to the CPU EP; `CRISPASR_ORT_FORCE_CPU=1` disables CUDA explicitly. The GGUF/ggml path is untouched — this is strictly additive and opt-in (`CRISPASR_USE_ONNXRUNTIME=ON`).

## Why

- The ggml ports of pyannote-seg and TitaNet are CPU-only, which makes diarization slow on long recordings (#326).
- CUDA builds historically crashed with the GGUF pyannote path (#98). Running the well-tested ONNX exports (sherpa-onnx's pyannote-segmentation-3-0, NVIDIA NeMo's TitaNet) through onnxruntime avoids maintaining a second implementation of those graphs in ggml.

## Changes

- `src/core/ort_session.h` — onnxruntime session factory; attaches the CUDA EP when the linked distribution exports it, falls back to CPU; `CRISPASR_ORT_FORCE_CPU` override.
- `src/pyannote_seg.cpp` — `.onnx` dispatch in `pyannote_seg_init` / `pyannote_seg_run`.
- `src/titanet.cpp` — same for TitaNet embeddings.
- `CMakeLists.txt` / `src/CMakeLists.txt` — `CRISPASR_USE_ONNXRUNTIME` option + `ONNXRUNTIME_ROOT`; link dirs/libs are PUBLIC because `crispasr-lib` is static and the final link (server/CLI executables) must see the onnxruntime paths.
- `tools/test_ort_poc.cpp` — harness comparing GGUF vs ONNX TitaNet embeddings and running the pyannote-segmentation ONNX on raw 16 kHz PCM.
- `README.md` — build documentation for the option.

## Follow-up fixes found during end-to-end verification

1. `resolve_pyannote_model` only accepted `.gguf` paths, so `--sherpa-segment-model <model.onnx>` made the native path unreachable and the pipeline silently fell back to the sherpa-onnx subprocess. It now accepts `.onnx`.
2. The spectral clusterer (`core_spectral::cluster_speakers`) collapsed to k=1 on short inputs (GMM/BIC degeneration with few, near-duplicate embeddings), merging distinct speakers into a single label. The distinct pyannote local tracks already on the segments (single full-audio forward pass via the #107 cache → locally consistent track indices) are now passed as `min_speakers`.
3. CMake: onnxruntime link dirs/libs must be PUBLIC (see above) — otherwise `crispasr-server` fails to link with `cannot find -lonnxruntime`.

## Verification (real GPU: onnxruntime-gpu 1.21.0, CUDA 12)

- Server binary built with `CRISPASR_USE_ONNXRUNTIME=ON`; both models log the CUDA EP at load:
  `pyannote_seg[ort]: .../model.onnx (CUDA EP)` and `titanet[ort]: .../nemo_en_titanet_large.onnx (CUDA EP)`.
- End-to-end `--diarize-method pyannote` on the sherpa-onnx multispeaker reference (2 speakers, 4 turns): output segments are labelled **A B A B** — the correct alternating pattern. Clusterer log: `n_emb=4 dim=192 -> k=2`.
- TitaNet ONNX vs GGUF equivalence on the same clips: cosine 0.71–0.72; cross-speaker separation 0.08–0.10.

## Note for users

Use onnxruntime **>= 1.20**: 1.19.x requires `libcudnn.so.8`, while current CUDA 12 images ship cuDNN 9, so the CUDA EP fails to attach and the models silently run on CPU.

## Related

#98 (pyannote diarization on CUDA builds), #326 (diarization too slow), #107 (full-audio pyannote cache).

## AI-assisted authorship

This pull request was prepared with the assistance of an AI coding agent (Hermes Agent, Nous Research). The implementation was verified end-to-end on real hardware (RTX 3060, onnxruntime-gpu 1.21.0, CUDA 12) before submission, and the PR was commissioned by the repository owner (tilllt).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,11 @@ option(CRISPASR_USE_SYSTEM_GGML "crispasr: use system-installed GGML library" OF
 option(CRISPASR_PORTABLE_CPU
        "crispasr: compile bundled ggml-cpu for generic x86-64 (no optional ISA)"
        OFF)
+option(CRISPASR_USE_ONNXRUNTIME
+       "crispasr: enable onnxruntime inference for pyannote-seg/TitaNet (model paths ending in .onnx); GPU via CUDA EP with CPU fallback"
+       OFF)
+set(ONNXRUNTIME_ROOT "" CACHE PATH
+       "crispasr: path to the onnxruntime C++ distribution (contains include/ and lib/); required when CRISPASR_USE_ONNXRUNTIME=ON")
 
 # sanitizers
 option(CRISPASR_SANITIZE_THREAD    "crispasr: enable thread sanitizer"    OFF)
--- a/README.md
+++ b/README.md
@@ -577,6 +577,36 @@ cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_METAL=ON    # Apple Silicon
 cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_VULKAN=ON   # cross-vendor
 ```
 
+### Optional: onnxruntime for diarization models (`.onnx` paths)
+
+The pyannote segmentation and TitaNet speaker-embedding nets used by
+`--diarize-method pyannote` normally run on the CPU-only ggml runtime.
+When built with `CRISPASR_USE_ONNXRUNTIME=ON`, any model path ending in
+`.onnx` is executed by [onnxruntime](https://onnxruntime.ai/) instead —
+the CUDA execution provider is attached automatically when the linked
+distribution provides it (use `onnxruntime-gpu`), with a silent fallback
+to the CPU EP; `CRISPASR_ORT_FORCE_CPU=1` disables CUDA explicitly.
+The GGUF/ggml path is untouched.
+
+```bash
+cmake -B build -DCMAKE_BUILD_TYPE=Release \
+  -DCRISPASR_USE_ONNXRUNTIME=ON \
+  -DONNXRUNTIME_ROOT=/path/to/onnxruntime-linux-x64-gpu-1.21.0
+cmake --build build -j$(nproc)
+```
+
+Use **onnxruntime ≥ 1.20**: 1.19.x requires `libcudnn.so.8`, while current
+CUDA 12 images ship cuDNN 9 — the mismatch makes the CUDA EP fail to attach
+and the models silently run on CPU. Known-good exports for this path:
+
+- `sherpa-onnx-pyannote-segmentation-3-0/model.onnx` (from the
+  [sherpa-onnx speaker-diarization release](https://github.com/k2-fsa/sherpa-onnx/releases/tag/speaker-diarization-models))
+- `nemo_en_titanet_large.onnx` (NVIDIA NeMo, exported via `torch.onnx.export`)
+
+`tools/test_ort_poc.cpp` (built as `build/bin/test-ort-poc`) compares the
+GGUF vs ONNX TitaNet embeddings on two speaker WAVs and runs the pyannote
+segmentation ONNX on raw 16 kHz PCM.
+
 **See [`docs/install.md`](docs/install.md)** for the full guide:
 all GPU backends (CUDA / Metal / Vulkan / MUSA / SYCL), Windows
 convenience scripts, ffmpeg ingestion, optional BLAS, glibc notes,
--- a/examples/cli/crispasr_diarize_cli.cpp
+++ b/examples/cli/crispasr_diarize_cli.cpp
@@ -17,6 +17,8 @@
 #include "speaker_db.h"
 #include "whisper_params.h"
 
+#include <set>
+
 #include "core/spectral_diarize.h"
 
 #include <algorithm>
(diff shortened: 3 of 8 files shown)


CrispStrobe:
Thank you for this — the end-to-end verification on real CUDA hardware is more than most PRs arrive with, and two of the three follow-up fixes you found are real bugs I have now reproduced independently. I am not going to take the onnxruntime dependency though, and I want to show you the measurements rather than just assert a preference, because they contradict both premises the PR rests on.

## The 0.71 cosine is a bug in the ORT path, not in ours

> TitaNet ONNX vs GGUF equivalence on the same clips: cosine 0.71–0.72

That is not equivalence — two implementations of one model on one input should agree to ~0.999, and 0.71 is close to the signature of a real bug we fixed in the sibling CAMPPlus embedder (#334, cos 0.737, a dropped fused-BatchNorm bias).

The upstream ONNX takes **mel features**, not PCM, so the two halves can be separated. Feeding our mel to `nemo_en_titanet_large.onnx`:

```
an255-fash-b:  ggml vs onnx (same mel) = 0.999996
cen7-fash-b:   ggml vs onnx (same mel) = 0.999996
```

And end-to-end, our embedding against the ONNX driven by an independent torch/librosa reimplementation of NeMo's `AudioToMelSpectrogramPreprocessor`, on NVIDIA's own two reference clips:

```
an255-fash-b:  0.9910        cen7-fash-b:  0.9964
```

So our network is right to six decimals and our front-end is right to three. The 0.71 is in the ORT integration — most likely its mel front-end, since the ONNX cannot compute one for you. Worth chasing on your side, because it means the CUDA path you measured was producing subtly wrong embeddings.

I added `CRISPASR_TITANET_DUMP_MEL` (7c30d09e) so this is checkable by anyone: dump our mel, feed it to any upstream export, and you get the network-only number. A single end-to-end cosine cannot tell you which half is wrong.

## We are already faster than onnxruntime

> The ggml ports of pyannote-seg and TitaNet are CPU-only, which makes diarization slow

Same 2 s segment, one M1, interleaved:

| | ms/segment |
|---|---|
| **ours (default path)** — *including* the mel | **71.7** |
| onnxruntime CoreML EP | 116.2 |
| onnxruntime CPU EP, 4 threads | 112–126 |

We are ~1.6× faster than the best provider onnxruntime offers on this machine, while doing strictly more work (ORT is handed precomputed mels).

## GPU, without the dependency — and it lost

You were right that the GPU was sitting idle: `titanet_init` already created a GPU backend and loaded the weights onto it, then computed on a separate CPU backend. `CRISPASR_TITANET_GPU=1` now points compute at it, exact to cosine 1.000000 against the CPU paths.

It is **opt-in, because it is slower where it counts**. Micro-benchmark says 31.9 ms (2.2× the default), but diarization embeds one segment per call at *variable* lengths — so each call reshapes the graph and the allocator re-reserves — and the embedder now runs several workers that contend for one GPU. End-to-end, 600 s clip, 47 segments:

```
workers=4, legacy   9994 ms   <- default
workers=1, legacy  12673 ms
workers=1, GPU     15275 ms
workers=4, GPU     49866 ms   <- 6x slower
```

That is the part I would not have believed without measuring, and it is why routing to a GPU runtime is not automatically a win for this workload. Length-bucketing so the graph shape stops changing is the work that would make GPU win generally; that is tracked separately.

## What I would like to take from this PR

Your finding **#2 is a genuine bug** and I reproduced it with no ONNX involved — on `samples/multispeaker.wav` (2 speakers) the default path reports **one**:

```
pyannote + titanet : n_emb=5 -> k=1 (gmm_bic)   -> 1 speaker   WRONG
foxnose            :                            -> 2 speakers  correct
```

Your fix is well reasoned, including the "distinct local tracks, not max+1" subtlety. **Could you split it into its own PR?** It stands entirely on its own, it fixes a real accuracy bug in the default configuration, and it should land with your name on it rather than me lifting it out.

Finding #1 (`resolve_pyannote_model` rejecting non-`.gguf`) becomes moot without the ORT path, and #3 is ORT-specific.

Two other things worth your time if you want to keep going here: **#98** (CUDA crashes with the GGUF pyannote path) is a bug we would want fixed rather than routed around — going via onnxruntime leaves it in place for everyone not building with ORT. And pyannote-seg is SincNet → 3× biLSTM, so its recurrence is sequential and a GPU cannot parallelise the dominant part; the win there was chunked inference, which landed in v0.8.28 (32.8 s → 5.5 s).

Closing this one as not-merging, with genuine thanks — the clusterer bug alone was worth the review, and the 0.71 number sent me to a measurement I should have had already.

Issue #375, opened 2026-08-19
Regression in speech recognition with canary


cdoepmann:
I've been running an older version of CrispASR for some time, being very happy with its ability to run canary. Unfortunately, since I bumped CrispASR it to a more recent version, I am now experiencing a clear degradation in recognition quality. Every now and then sentences are interrupted and words/phrases are repeated a few times before the recognition continues. The issue appears with all quantizations and also running on CUDA as well as on CPU. 

I tried to git bisect the issue that introduced this problematic behavior. It seems to have been introduced between 282e5d0b98f90cff08d7ad75bfa454034fa011c4 (good) and c08f7a524eeb0e24b4e6abba061629cf42db0b06 (bad). I cannot specify the commit more precisely because compilation fails in that range due to other reasons.

I would be very grateful for having this fixed.

CrispStrobe:
Thanks for the careful bisect — the window you gave is genuinely useful, and I've
been able to rule several things out with it. But I have to be straight with you:
**I cannot reproduce the regression**, and I don't want to ship a fix for it
blind.

## What I tested

Built your last-good commit `282e5d0b` and current `main`, ran both on the same
audio with `canary-1b-v2-q4_k` on CPU:

| audio | `282e5d0b` vs `main` |
|---|---|
| 132 s, 16 chunk seams | **byte-identical** (`7cce22b7…`) |
| 594 s natural German, ~74 seams | **byte-identical** (`58346dcd…`) |

Zero difference across both. I do see seam artifacts — repeated phrases like
`"ask not Ask not"` and sentences broken mid-clause — but they are **equally
present on your good commit**, so they are a pre-existing defect rather than what
changed for you.

## What that rules out

- **ggml**: the submodule pointer is identical at both endpoints
  (`0714117daca2471b00e09554c7eaa74a06b0b2c5`), so nothing in the ggml bump family
  is involved.
- The one commit in the window that touches canary sources (`73bb9b2f`, an
  encoder-graph use-after-free fix) does not change output on either of my files.

## What would let me fix it

Any one of these moves this forward a lot; the first is worth the most:

1. **A sample of the audio that misbehaves** — even 30–60 s that shows one
   repetition. Language, sample rate and channel count matter. If it's not
   shareable, a synthetic clip in the same language/recording conditions that
   still shows it is fine.
2. **The exact model file** (which GGUF, which quant) and **the exact command**
   you run, including any env vars.
[...]

Issue #365, opened 2026-08-17
ASR - Canary backend : some chunks of segments get duplicated or overlapped in the output


AmineI:
Hello ! The segment overlap symptom is quite clear in the below example output. Even the timestamps go back in time.

The fact that the overlap is textually different (grammar) shows that this is not just a duplicated line in the output, but the model itself redoing the work in another context.

```
00:20:24,020 --> 00:20:26,180
Ça c'est quand elle s'est rendue compte
362
00:20:26,180 --> 00:20:27,140
qu'elle n'avait pas dormi.
363
00:20:27,060 --> 00:20:28,340
J'arrive sur le...
364
00:20:26,820 --> 00:20:28,020
rendu compte qu'elle n'avait pas dormi.
365
00:20:27,940 --> 00:20:29,300
J'arrive sur le salut monsieur.
```

The vad segments seem properly separated 
```

whisper_vad_segments_from_probs: VAD segment 195: start = 1214.43, end = 1218.91 (duration: 4.48)
whisper_vad_segments_from_probs: VAD segment 196: start = 1220.35, end = 1222.33 (duration: 1.98)
whisper_vad_segments_from_probs: VAD segment 197: start = 1225.35, end = 1227.58 (duration: 2.23)
whisper_vad_segments_from_probs: VAD segment 198: start = 1227.84, end = 1233.82 (duration: 5.98)
whisper_vad_segments_from_probs: VAD segment 199: start = 1234.37, end = 1235.58 (duration: 1.21)
whisper_vad_segments_from_probs: VAD segment 200: start = 1238.31, end = 1245.85 (duration: 7.54)
whisper_vad_segments_from_probs: VAD segment 201: start = 1246.53, end = 1252.09 (duration: 5.56)
[...]
whisper_vad_segments_from_probs: VAD segment 361: start = 1905.73, end = 1906.24 (duration: 0.51)
whisper_vad_segments_from_probs: VAD segment 362: start = 1906.59, end = 1908.06 (duration: 1.47)
whisper_vad_segments_from_probs: VAD segment 363: start = 1908.29, end = 1912.13 (duration: 3.84)
whisper_vad_segments_from_probs: VAD segment 364: start = 1914.88, end = 1916.29 (duration: 1.41)
whisper_vad_segments_from_probs: VAD segment 365: start = 1919.07, end = 1923.36 (duration: 4.29)
[...]

PR #376, opened 2026-08-19
fix(canary): prevent repeated streamed overlap in #375

CrispStrobe:
## What changed

Canary streamed decoding now removes a conservative multi-token prefix when its token end times are already covered by the previous chunk. This catches reworded overlap that token-ID LCS cannot match, without enabling the broader fuzzy seam dedup or reverting the 73bb9b2f scheduler UAF fix. Single early tokens remain untouched.

## Validation

- `cmake --build /tmp/crispasr-build-375b --target test-asr-overlap-trim canary -j2`
- `test-asr-overlap-trim` (19 assertions, 8 cases)

The reporter did not attach audio/model details, so live reproduction on the HP Z800 input is still needed.

### DIFF
--- a/src/canary.cpp
+++ b/src/canary.cpp
@@ -1684,6 +1684,25 @@ extern "C" struct canary_result* canary_transcribe_streamed(struct canary_contex
             }
         }
 
+        // #375: after the encoder-graph lifetime fix (73bb9b2f), the AED can
+        // choose different wording for the overlap on adjacent chunks.  The
+        // token-id LCS above then finds no match and the re-transcribed prefix
+        // is appended verbatim, producing repeated phrases (and timestamps
+        // that move backwards).  Apply the conservative time signal here:
+        // only trim a prefix of at least two complete tokens whose end is
+        // already at or before the last accepted token.  A one-token prefix is
+        // intentionally retained; boundary timing can put a genuine first
+        // word slightly early, and this avoids the false-positive seen in
+        // #365 ("Many").  The fuzzy word matcher remains opt-in for cases
+        // where timing alone is insufficient.
+        if (chunks_processed > 0 && !all_tokens.empty() && part->n_tokens > n_skip) {
+            const int64_t accepted_end_cs = all_tokens.back().t1;
+            const int covered = core_overlap_trim::leading_covered_multi(
+                part->n_tokens, accepted_end_cs, [&](int i) { return part->tokens[i].t1; });
+            if (covered > 0)
+                n_skip = std::max(n_skip, covered);
+        }
+
         // #365: fuzzy seam dedup, UNDER the LCS result.
         //
         // The LCS above compares token ids, so it only fires when the decoder
--- a/src/core/asr_overlap_trim.h
+++ b/src/core/asr_overlap_trim.h
@@ -63,6 +63,14 @@ template <typename F> inline int leading_covered(int n, int64_t accepted_end_cs,
     return skip;
 }
 
+// Conservative streamed-seam variant. A single early token can be a genuine
+// first word whose decoder timestamp straddles the seam; require at least two
+// covered tokens before treating the prefix as re-transcribed overlap.
+template <typename F> inline int leading_covered_multi(int n, int64_t accepted_end_cs, F t1_of) {
+    const int skip = leading_covered(n, accepted_end_cs, t1_of);
+    return skip >= 2 && skip < n ? skip : 0;
+}
+
 
 // ---------------------------------------------------------------------------
 // Fuzzy seam dedup
--- a/tests/test-asr-overlap-trim.cpp
+++ b/tests/test-asr-overlap-trim.cpp
@@ -80,6 +80,16 @@ TEST_CASE("overlap trim: an entirely covered chunk reports every token", "[unit]
     REQUIRE(trim(next_chunk, 99999) == 2);
 }
 
+TEST_CASE("overlap trim: conservative seam rule requires two covered tokens", "[unit][overlap][issue-375]") {
+    const std::vector<Tok> one_early = {{" Many", 100, 200}, {" people", 200, 400}};
+    REQUIRE(core_overlap_trim::leading_covered_multi(
+                (int)one_early.size(), 250, [&](int i) { return one_early[(size_t)i].t1; }) == 0);
+
+    const std::vector<Tok> repeated = {{" rendu", 100, 200}, {" compte", 200, 250}, {" real", 250, 400}};
+    REQUIRE(core_overlap_trim::leading_covered_multi(
+                (int)repeated.size(), 250, [&](int i) { return repeated[(size_t)i].t1; }) == 2);
+}
+
 TEST_CASE("overlap trim: scanning stops at the first uncovered token", "[unit][overlap][issue-365]") {
     // A decoder can emit a stray token whose timing sits behind its neighbours.
     // Scanning the whole chunk and counting every covered token would drop real


CrispStrobe:
Reviewed with a live reproduction, which is the thing this PR says it still
needs. **I do not think this should merge as written** — measured, it does not
fix the symptom, and it makes the output worse in a new way.

## Reproduction

`canary-1b-v2-q4_k`, CPU, `jfk_x12.wav` (132 s = the JFK sentence 12×). A
repeating source is a good probe here because the correct answer is exactly 12
copies, so anything extra is duplication rather than a judgement call.

`main` reproduces the reporter's description precisely:

```
... ask what you can do. for your country. And so, my fellow Americans, ask not Ask not what your country ...
                       ^^^^ sentence interrupted                            ^^^^^^^^^^^^ phrase repeated
```

Both halves of #375 are present: "sentences are interrupted" **and**
"words/phrases are repeated".

## What this PR does to it

The entire effect across 132 s is two deleted tokens:

```
main:  ask not  Ask not  what your country ...
#376:  ask not      not  what your country ...
```

It removes `"Ask"` and leaves `"not"`. The duplication survives as
`"ask not not"` — still duplicated, and now ungrammatical. The two
interrupted-sentence artifacts are untouched.

So on the only end-to-end evidence available, the PR converts one visible defect
into a different visible defect and fixes nothing else.

## The mechanism it declines to use already fixes this

`CRISPASR_CANARY_SEAM_DEDUP=1` — the #365 fuzzy matcher already in the tree —
removes **both** tokens on the same file:

| build | at the seam |
|---|---|
| `main` | `ask not Ask not what your country…` |
| this PR | `ask not not what your country…` |
| `CRISPASR_CANARY_SEAM_DEDUP=1` | `ask not what your country…` ✅ |

That is not a coincidence: the fuzzy matcher's own comment describes this exact
failure — "given a different amount of right-context an AED re-words the overlap
instead … the LCS matches nothing". `"ask"` vs `"Ask"` is a capitalisation
rewording, which is precisely what token-id LCS cannot see and normalised word
matching can.

## Two more concerns with the approach

**It reintroduces the thing that was already measured as harmful.** The comment
directly above the insertion point says: *"Deliberately NOT a plain time floor.
Trimming the overlap by timestamp alone was measured removing real speech — the
leading 'Many' … since an AED's boundary timings overshoot."* This PR adds a
plain time floor with `>= 2 tokens` as the mitigation. The 2-token threshold is
tuned to one anecdote, and the failure mode it guards against is silent deletion
of real speech — which is strictly worse than duplication, because the user can
see duplication and cannot see a dropped phrase.

**`leading_covered_multi` excludes the clearest case.** `skip >= 2 && skip < n`
returns 0 when the whole chunk is covered — i.e. a fully re-transcribed chunk,
the most certain duplicate there is, is deliberately not trimmed, while partial
overlaps are. If that is intentional safety, it is worth saying so; as written
the rule is hardest on the cases it is least sure about.

## On the diagnosis

"The only Canary change in the window is 73bb9b2f" is true, but it is a
**file-scoped** search over a 41-commit window that the reporter explicitly could
not narrow ("compilation fails in that range"). By construction that method
cannot see cross-cutting changes, and this window has six commits touching the
audio input path (`f3d82d30` routes CLI + stereo decode through glint,
`3739e70d` clamps file parsers). I checked and there is **no ggml submodule bump**
in the range, so that one is out — but 73bb9b2f was *selected*, not bisected, and
the PR presents it as identified.

For what it's worth I think the causal story is still plausible in one direction:
before 73bb9b2f the cached encoder graph could hand adjacent chunks stale
buffers, which would make them decode *identically* and let the token-id LCS
match. Fixing the UAF made each chunk genuinely independent, so the AED rewords
and the LCS stops firing. That argues for a rewording-tolerant seam merge — the
fuzzy matcher — rather than a time floor.

## Suggested path

1. Don't merge as-is.
2. Reproduce first — `jfk_x12.wav` above works, and the reporter can also run
   `CRISPASR_CANARY_STREAM_THRESHOLD_S=99999` to force single-pass; if the
   duplication vanishes, the seam hypothesis is confirmed on *their* audio.
3. The real decision is whether `CRISPASR_CANARY_SEAM_DEDUP` should default on
   for canary. It fixes this repro cleanly; it was gated because it dropped a
   leading "Many" on a 600 s clip. That is a corpus question, not a one-clip
   question — the honest next step is running both settings over the regression
   audio and counting insertions vs deletions, rather than adding a third
   heuristic beside the two we already have.
4. Note that neither mechanism touches the *interrupted sentence* half
   (`"you can do. for your country."`), which is arguably the larger part of what
   the reporter is complaining about. That looks like splice punctuation, not
   dedup, and should be tracked separately rather than folded into this.

CrispStrobe:
**Correcting my own review above — its central evidence was invalid, and the
conclusion I drew from it does not stand.**

I claimed "`main` reproduces the reporter's description precisely" on
`jfk_x12.wav`. I never checked that the same artifacts are absent on the
reporter's GOOD commit. They are not. Built `282e5d0b` (their last-good) and ran
the identical file and model:

```
282e5d0b : 7cce22b79c7cae62f05591caa512fe6c
main     : 7cce22b79c7cae62f05591caa512fe6c
```

Byte-identical — including all three artifacts I quoted (`ask not Ask not`,
`you can do. for your country.`, `for yourself. your country.`). So those are a
**pre-existing seam defect present in the reporter's own working build**, not the
regression. My "reproduction" reproduced the wrong thing.

Repeated on 594 s of natural German long-form (`de/fleurs_600s.wav`, ~74 seams,
which is much closer to the reporter's likely use than a repeating clip):

```
282e5d0b : 58346dcdf82aa6c1eda898cc8f4f75b7
main     : 58346dcdf82aa6c1eda898cc8f4f75b7
```

Also byte-identical. `canary-1b-v2-q4_k`, CPU, `-l de`.

## What this means for the PR

My measured objections to the *implementation* still hold on their own terms —
turning `ask not Ask not` into `ask not not` is worse than leaving it, and the
`skip >= 2 && skip < n` rule excludes the fully-covered case. But my headline
("does not fix the symptom") was measured against the wrong symptom, so treat it
as a critique of the heuristic, not as a verdict on #375.

The real problem is the same for the PR and for my review: **nobody has
reproduced the regression.** Two long-form probes across the exact bisect
endpoints show zero difference, so neither this change nor my proposed
alternative (`CRISPASR_CANARY_SEAM_DEDUP=1`) can be validated as fixing anything.
Merging a decode-path heuristic that silently deletes tokens, on that basis,
risks trading an unreproduced bug for a real one.

I've asked the reporter for what's needed on #375. Until that lands I'd hold this.

## What is still solid

- ggml did not move in the window — the submodule pointer is
  `0714117daca2471b00e09554c7eaa74a06b0b2c5` at **both** endpoints, so that whole
  class is out.
- `73bb9b2f` is the only commit touching canary sources in the window, but it was
  *selected* by file-path filter, not bisected; the window also contains six
  commits touching the audio input path.
- The pre-existing seam artifacts above are worth their own issue regardless of
  #375 — they are real, reproducible on demand, and `CRISPASR_CANARY_SEAM_DEDUP=1`
  cleanly fixes the duplication half of them.

CrispStrobe:
Status update: the #375 regression itself turned out to be the glint AAC-LC decoder (KBD windows discarded + TNS mis-decode → ~17 dB SNR on every real-world .aac), fixed on main in 0e5d1344 — details in https://github.com/CrispStrobe/CrispASR/issues/375#issuecomment-5346753096. The seam duplication this PR targets is real but pre-existing (present on the reporter's good commit too) and is now tracked in #378, including the comparison against the gated #365 fuzzy matcher (`CRISPASR_CANARY_SEAM_DEDUP=1`) that the corpus decision should be based on. Holding this PR pending that comparison — the earlier review concerns (deletes one token of a two-token duplicate; `leading_covered_multi` declines the fully-covered case) still stand.

CrispStrobe:
Superseded — #375's root cause was the glint AAC decoder (fixed in 0e5d1344), and the seam handling is being aligned with the NeMo blueprint directly.

Issue #208, opened 2026-06-29
Session/Rust API has no chunked Parakeet path — long audio runs one full-length FastConformer encode (O(T²) attention), effectively hangs


davideme:
## Summary

The session-level transcribe API (`crispasr_session_transcribe[_lang]`) and the
Rust `crispasr::Session` binding always run the **Parakeet** backend through the
**unchunked, full-length** encoder (`parakeet_transcribe_ex` → `parakeet_encode_mel`).
For long audio this builds one enormous FastConformer graph whose global
relative-position self-attention is O(T²) in the encoder time dimension
(T_enc = T_mel / 8). On the Metal backend the GPU then grinds for minutes on a
single `ggml_backend_sched_graph_compute`, which from the host side is
indistinguishable from a hang.

The library *already has* a bounded encoder — `parakeet_encode_chunked`
(`src/parakeet.cpp:2890`, used by `parakeet_transcribe_chunked` / `parakeet_transcribe_streaming`)
— but it is **not exposed** through the session C-API or the Rust binding, so
downstream batch callers can't reach it.

Observed on rev `8f1218141b792b8868861c1af17ba1e361b05dc0` (dylib `0.10.2`),
Apple M1, macOS 26.5.1, Metal backend, Parakeet-TDT.

## Where the path is fixed to the unchunked encoder

- `crispasr_session_transcribe_lang` → `transcribe_single`, parakeet branch:
  `src/crispasr_c_api.cpp:3477-3478`
  ```cpp
  if (s->backend == "parakeet" && s->parakeet_ctx) {
      parakeet_result* pr = parakeet_transcribe_ex(s->parakeet_ctx, pcm, n_samples, 0);
  ```
- `parakeet_transcribe_ex` encodes the whole buffer in one shot:
  `src/parakeet.cpp:3189`
  ```cpp
  enc = parakeet_encode_mel(ctx, mel.data(), (int)ctx->model.hparams.n_mels, T_mel, &T_enc);
  ```
- The bounded alternative that is **not** wired into the session path:
  `parakeet_encode_chunked` (`src/parakeet.cpp:2890`), `parakeet_transcribe_chunked`
  (`:2960`), `parakeet_transcribe_streaming` (`:3070`). None are `CA_EXPORT`ed.

## VAD path is not a substitute

`crispasr_session_transcribe_vad_lang` only strips silence — when there are
multiple speech slices it **stitches them into one buffer and transcribes once**
[...]

PR #386, opened 2026-08-24
fix(#385): fire session progress callback on unified long-form dispatch

### LINKED ISSUE #385: Session progress callback never fires on long-form Parakeet runs, report no progress.
## Summary

`crispasr_session_set_progress_callback` + `crispasr_session_transcribe_chunked[_lang]` stopped delivering per-window progress. The callback is registered, the audio is chunked and decoded window by window, but the unified orchestrator never invokes `s->progress_cb` — so a caller driving a progress bar sees 0% until the call returns.

This is the progress-reporting sibling of #350: same entry point, same dispatch change, different symptom. #350's fix (v0.8.29) restored the *chunking*; the *reporting* is still gone.

## Where the wiring stops (v0.8.29, `3721d402`)

The callback is stored per-session in `crispasr_session_set_progress_callback` (`src/crispasr_c_api.cpp:5107`). It is invoked on exactly two paths:

- the legacy inline path, `parakeet_session_chunked_merge` — fires once per finished window with monotonically increasing `(end, n_samples)` (`src/crispasr_c_api.cpp:4974`), reachable only with `CRISPASR_SESSION_UNIFIED_DISPATCH=0`;
- the JA sliced path — fires per slice (`src/crispasr_c_api.cpp:5471`).

The default non-JA route, however, hands off to the shared orchestrator at `src/crispasr_c_api.cpp:5354`:

```cpp
for (auto& ps : parakeet_transcribe_segments(s->parakeet_ctx, pcm, n_samples, 0, is_ja, oo)) {
```

Neither `parakeet_orchestrate_opts` nor anything in `src/parakeet_orchestrate.cpp` carries or calls a progress callback — `grep progress src/parakeet_orchestrate.cpp` comes back empty. The module-level `g_progress` atomic (Dart polling) is equally untouched on this path.

## Repro

Any non-JA Parakeet model, audio long enough to chunk:

```rust
session.transcribe_chunked_with_progress(pcm, 0, -1, None, |processed, total| {
    eprintln!("{processed}/{total}");
})
```

davideme:
Fixes #385.

## Problem

The #350 unified-dispatch fix (v0.8.29) restored the *chunking* through the shared orchestrator, but the *reporting* stayed broken: `parakeet_transcribe_segments` carried no progress hook, so a caller of `crispasr_session_set_progress_callback` + `crispasr_session_transcribe_chunked[_lang]` saw 0 % until the call returned, and the pollable `g_progress` atomic (Dart FFI) stayed at idle too.

## Fix

Hoist the #208 progress contract into the orchestrator:

- `parakeet_orchestrate_opts` gains `progress_cb`/`progress_ud`, fired once per finished LONGFORM window with `(samples processed so far, total samples)` — monotonically non-decreasing, reaching `total` on the last window. Both the serial and pipelined loops fire it, **including a window whose encode failed**, so a progress bar cannot stall short of 100 %.
- The single-decode routes (SINGLE_PASS / STREAMED / CHUNK_SEGMENTED) have no observable windows and stay silent, per the session header's documented contract; the #350 gap-fill repair runs after 100 % and stays silent too.
- The session's unified branch wraps `s->progress_cb` and the `g_progress` mirror into that hook, in lockstep as documented, bracketed by the same started(0)/idle(-1) stores the legacy inline path used.
- Header doc: noted the 0.8.24–0.8.29 gap on `crispasr_progress_callback`, mirroring the existing #350 note.

## Verification

Live on `parakeet-tdt-0.6b-v3-q4_k` (M1 Pro, Metal):

- 66 s chunked run through the session ABI: 3 windows fire (`400800 → 880800 → 1056000/1056000`), poller sees 37/83/100 %, atomic resets to −1 after — previously zero fires.
- Legacy A/B path (`CRISPASR_SESSION_UNIFIED_DISPATCH=0`) unchanged: 5 fires, same contract.
- New regression case in `test_parakeet_longform_live.cpp` (`[parakeet-longform]`, model-gated, SKIPs cleanly): ≥ 2 fires on a 66 s chunked run, monotonic, ends at `(total, total)`, atomic advances during the run and returns to idle after. All 3 cases in that binary pass, as do `test-parakeet-strategy` (56 assertions) and `test-issue-356-gap-fill-overlap` (90 assertions).

### DIFF
--- a/include/crispasr_session.h
+++ b/include/crispasr_session.h
@@ -71,6 +71,12 @@ CRISPASR_SESSION_API void crispasr_reset_progress(void);
 // Single-pass and non-Parakeet backends do not fire it. The module-level
 // atomic (crispasr_get_progress) is updated in lockstep, so pure pollers
 // (e.g. Dart FFI) get chunked progress without registering a callback.
+//
+// Issue #385: between the unified-dispatch switch (0.8.24) and 0.8.29 the
+// default non-JA Parakeet path ran through a shared orchestrator with no
+// progress hook, so neither the callback nor the atomic moved until the
+// call returned. The hook now lives in the orchestrator itself (one fire
+// per long-form window), so the contract holds on every dispatch path.
 typedef void (*crispasr_progress_callback)(int processed, int total, void* user_data);
 
 // Register (or clear, with cb == NULL) the session progress callback.
--- a/src/crispasr_c_api.cpp
+++ b/src/crispasr_c_api.cpp
@@ -5556,8 +5556,12 @@ static crispasr_session_result* transcribe_single(crispasr_session* s, const flo
                 }
                 for (auto& w : sw)
                     words.push_back(std::move(w));
-                if (s->progress_cb)
-                    s->progress_cb((int)si + 1, (int)slices.size(), s->progress_ud);
+                // Issue #385: report through the shared helper so the pollable
+                // g_progress atomic moves in lockstep with the callback — this
+                // path used to fire the callback alone, leaving Dart-style
+                // pollers at 0 % for the whole JA run. Units are slices, not
+                // samples (pre-existing on this path).
+                session_report_progress(s, (int)si + 1, (int)slices.size());
             }
             std::sort(
                 words.begin(), words.end(),
--- a/src/crispasr_c_api.cpp
+++ b/src/crispasr_c_api.cpp
@@ -4944,6 +4944,23 @@ CA_EXPORT crispasr_session_result* crispasr_session_transcribe_lang(crispasr_ses
 // session path bounds the FastConformer encode on long audio instead of
 // building one O(T^2) full-length graph.
 
+// The one definition of the #208 progress contract: mirror the pollable
+// g_progress atomic, then the per-session callback, in lockstep — every
+// long-form path reports through here so the two cannot drift apart (#385
+// was exactly that drift: a path that updated neither).
+static void session_report_progress(crispasr_session* s, int processed, int total) {
+    g_progress.store(total > 0 ? (int)((int64_t)processed * 100 / total) : 0, std::memory_order_relaxed);
+    if (s && s->progress_cb)
+        s->progress_cb(processed, total, s->progress_ud);
+}
+
+// Issue #208: started(0) / idle(-1) bracket around a long-form run, RAII so
+// no exit — early return or failure — can leak a stale percentage to pollers.
+struct scoped_session_progress {
+    scoped_session_progress() { g_progress.store(0, std::memory_order_relaxed); }
+    ~scoped_session_progress() { g_progress.store(-1, std::memory_order_relaxed); }
+};
+
 // Normalize a word for boundary-dedup comparison: lowercase + drop ASCII
 // punctuation. Non-ASCII bytes (JA / accented text) are kept verbatim.
 static std::string parakeet_norm_word(const char* s) {
@@ -4984,7 +5001,7 @@ static std::string parakeet_norm_word(const char* s) {
 static void parakeet_session_chunked_merge(parakeet_context* ctx, const float* samples, int n_samples,
                                            int chunk_samples, int overlap_samples,
                                            std::vector<crispasr_session_seg>& out,
-                                           crispasr_progress_callback prog_cb = nullptr, void* prog_ud = nullptr) {
+                                           crispasr_session* sess = nullptr) {
     const int SR = 16000;
     if (chunk_samples < SR)
         chunk_samples = SR; // 1 s floor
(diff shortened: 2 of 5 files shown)


davideme:
Two follow-up commits pushed:

- **`bee72966` refactor: one session progress definition, `std::function` hook.** `parakeet_orchestrate_opts` now carries `std::function<void(int,int)> on_progress` (the header is C++-only), so the session call site is a plain capturing lambda — the stack `prog_ctx` struct is gone. `session_report_progress()` is the single definition of the #208 contract (g_progress mirror, then callback, in lockstep), used by both the legacy chunked-merge path and the unified branch, and `scoped_session_progress` (RAII) replaces the four hand-written started(0)/idle(−1) store pairs so no early return can leak a stale percentage. Behavior-preserving: re-verified live on both dispatch paths (unified 3 fires, legacy A/B 5 fires, contract intact).

- **`b16e18df` fix: mirror g_progress on the JA sliced path too.** Found while unifying: the JA sliced path fired the callback but never touched the pollable atomic, leaving Dart-style pollers at 0 % for the whole JA run — the same lockstep drift as #385, in miniature. Now routed through the shared helper. Units stay slices (pre-existing on this path). Compile-verified; no JA model locally for a live run, but the helper itself is live-tested by the non-JA paths.

davideme:
CI triage:

- **clang-format** — real, mine: the reworked `parakeet_session_chunked_merge` signature and two spots in the test file. Fixed in `a5b2e336`.
- **linux-unit / linux-x86_64** — pre-existing on `main`, not from this PR: both jobs fail identically on main's current head `7ba3dcbe` (CI run 32739887612) with the same `confucius4-tts` wiring gaps (`confucius4_tts_init_from_file` absent from the shipped library; `test-arch-backend-map` "every emitted backend name is a name a surface can open"). Nothing in this PR touches backend wiring.

Also rebased the branch onto current `main` so the PR no longer carries two unrelated fork-sync commits (a glint submodule bump and a merge commit) — the diff is now exactly the four #385 commits. Re-verified after rebase: library builds clean, `test-parakeet-strategy` and the #385 live regression case pass, `clang-format --dry-run --Werror` clean on every touched file.

davideme:
@CrispStrobe [linux-x86_64](https://github.com/CrispStrobe/CrispASR/actions/runs/32817198458/job/97707777520?pr=386#logs) is also failing in main, do want to fix it in this PR or another ?

CrispStrobe:
Thanks — merged. This landed on `main` as part of `1819c848`; GitHub didn't auto-close the PR because the merge came in via a local branch rather than by advancing this one, so closing it by hand.

Nice diagnosis. Hoisting the hook into the orchestrator rather than re-adding it at the session call site is the right shape, and `session_report_progress` as the single definition of the #208 contract plus the `scoped_session_progress` RAII bracket are both improvements on what was there before — the old inline site could divide by zero on `total == 0`, and an early return could leave a stale percentage behind. `lf_window.end` being the pre-extension end also makes monotonicity structural rather than incidental.

### Verification on our side (Linux/CPU, `parakeet-tdt-0.6b-v3-q4_k`)

- `test-parakeet-strategy` 56 assertions / 17 cases, `test-issue-356-gap-fill-overlap` 90 / 10 — matching your numbers.
- The new #385 case passes with the same window sequence you measured on M1 Metal: `400800 → 880800 → 1056000/1056000`. Reproducing that on a different platform and backend is a good sign for the plan being deterministic.
- Legacy A/B (`CRISPASR_SESSION_UNIFIED_DISPATCH=0`) passes, 15 assertions.
- All three cases of the long-form suite pass; the #350 guards still hold (`gap-fill recovered 94 word(s)`).
- **The JA sliced path (c37e1270), which you could only compile-verify:** we have a JA parakeet model here, so we drove it through the session ABI. It fires 4 slices with `g_progress` mirroring the callback exactly (25/50/75/100) and returns to idle. Your change is correct. (Caveat for the record: the audio was English — JA-ness comes from the model vocab, so the right route ran, but this proves the progress plumbing, not JA transcript quality.)

### Conflict

`docs/feature-matrix.md` only. Your regen at 8c1de441 predates a1e7e67a, which declared `CAP_VOICE_CLONING`/`CAP_TEMPERATURE` on confucius4-tts. Rather than pick a side we regenerated from the merged tree's own binary — that came out as exactly your regen plus main's confucius4 row, which also confirmed main had drifted independently of your PR (vibevoice/-bitnet/-1.5b no longer declare `CAP_TEMPERATURE`, the chatterbox family gained src/tgt-language, `cosyvoice3-tts-rl` was missing from the table entirely). 108 backends now.

### Follow-ups we added on top

- `CRISPASR_MODEL_PARAKEET` is now exported from `tests/env-live-tests.sh`. It never was, so both your new #385 guard and the existing #350 guards were silently SKIPping in a normal live run.

### Still open

`crispasr_session_transcribe_chunked(…, chunk_seconds > 0, …)` sets `chunk_seconds_explicit`, so `parakeet_pick_strategy` returns `CHUNK_SEGMENTED` and nothing fires at all — where the legacy inline path did report per chunk. The JA route under the chunked entry point has the same shape: `force_chunked` excludes the sliced branch, so it falls through to the streamed fallback, also silent. `parakeet_transcribe_streamed` does have an internal per-encoder-chunk loop, so both are fixable, but it needs a `parakeet.h` signature addition and its own live test — tracking separately rather than widening this one.

CrispStrobe:
Merged as part of `1819c848` on `main` — closing since GitHub didn't detect it (the merge came in via a local branch, not by advancing this head). All five commits verified as ancestors of `main`, CI green on the tip.

Issue #324, opened 2026-07-29
feat: FoxNoseTech-diarize


smoab:
A promising alternative to PyAnnotate. Worth considering to add it as an option for diarization. See: https://github.com/FoxNoseTech/diarize

CrispStrobe:
you can test this now.

smoab:
Just tested v0.8.25 CUDA image, there seems to be an issue with foxnose diarize , tried with multiple small audio files (~1-2 minutes) , large portions of the audio didn't get transcribed. This also happened with Pyannotate only with VAD enabled. If vad=false and diarize_method=pyannotate , the audio get "almost" full transcription , some parts where there are overlapping speakers/laughter didn't get transcribed. The backend used is cohere-ar, using the command:

```
ubuntu@ubuntu:~$ curl --fail-with-body --show-error   http://localhost:8000/v1/audio/transcriptions   -F 'file=@/mnt/c/Users/xxx/Downloads/podcast-sample-ar.mp3'   -F 'language=ar'   -F 'response_format=diarized_json'   -F 'vad=true'   -F 'diarize=true'   -F 'diarize_method=foxnose'   -F 'diarize_embedder=auto'
```

CrispStrobe:
Fixed on main in f86193ff — thanks, that was a real bug and a bad one.

The server (not the CLI) diarized by re-walking the merged segment list and wrote back only the segment count it started with, but the speaker-turn splitter *grows* that list — so every sub-segment past the first was dropped, i.e. exactly the text around each speaker change. Short VAD slices usually hold one segment, so a 3-way split lost 2/3 ("large portions"); a 30 s no-VAD chunk holds many, so only the straddling ones vanished — your "overlapping speakers/laughter" parts *are* the split ones. Measured on your exact request (foxnose+vad): 30 words before vs 74 after, against an undiarized ground truth of 74.

Separately, foxnose was clustering each VAD slice independently on the server (numbering restarted every few seconds); it now runs one global pass like the CLI.

Please re-test when you can. If text is still missing with `diarize=false&vad=true`, that would be a different issue — this fix only touches the diarize post-step.

CrispStrobe:
Closing this — the feature is in and the bug @smoab found is verified fixed.

**FoxNose diarization** is available as `--diarize-method foxnose` (and `diarize_method=foxnose` on the server). On `samples/multispeaker.wav` the CLI gives the correct alternating pattern:

```
[...]

PR #396, opened 2026-08-26
feat(diarize): forward FoxNose speaker turns across the C ABI (#395)

### LINKED ISSUE #395: FoxNose callers can't split a segment that spans two speakers
## Symptom

On a real 10-minute, three-speaker recording, long stretches of transcript that
contain **two different people** come back labelled as **one speaker**. Not a
clustering failure — FoxNose gets the speaker count right and finds the
boundaries. The boundaries are just unreachable from the C ABI, so the label
resolution can never be finer than the segment grid the caller happened to send
in.

## Why it happens

`apply_foxnose` labels a caller segment with the turn it overlaps **most**
(`src/crispasr_diarize.cpp:547-560`):

```cpp
for (const auto& t : res.turns) {
    const double ov = std::min(b, t.end) - std::max(a, t.start);
    if (ov > best) { best = ov; best_spk = t.speaker; }
}
segs[i].speaker = best_spk;
```

So any segment straddling a speaker change is silently awarded to the majority
speaker, with no signal that it happened. The minority speaker's words are gone.

That isn't a pathological input on our side. Callers who build their grid from
ASR word timings can't submit a per-word grid, because `core_foxnose` skips any
span shorter than `kMinSegmentSeconds = 0.4 s`
(`src/core/foxnose_pipeline.h:26`, enforced at `foxnose_pipeline.cpp:32`): we
measured 309 words producing only 83 embedding windows, and the clusterer then
read 2 speakers where 3 were present. Merging words into utterance-length spans
recovers ~10 DER points — and that merge is exactly what buries sub-segment
speaker changes. It's a genuine bind: coarse segments cluster well and label
coarsely, fine segments starve the embedder.

Concretely, from our recording:

```
47.61  48.09  "razón."     <- speaker A finishes
48.27  48.41  "No,"        <- speaker B starts, 0.18 s later
```

Any merge threshold loose enough to keep windows above 0.4 s also fuses those
two into one segment.

## The information already exists — it's discarded one line before the return

`CRISPASR_DIARIZE_DEBUG=1` on that same recording:

```
crispasr[diarize]: foxnose windows=493 skipped=... -> 3 speakers (fixed), 176 turns
```

176 audio-derived turns over 10 minutes is precisely the resolution the caller
grid can't express. And the C++ entry point already hands them out:

```cpp
// src/crispasr_diarize.h:107
bool crispasr_diarize_segments(..., std::vector<CrispasrDiarizeTurn>* out_turns = nullptr);
```

whose own doc comment describes this exact use case:

> `out_turns`, when non-null, receives the speaker turns the method derived from
> the audio (FoxNose only). Callers with word timestamps can use them to split a
> segment that spans several speakers — labelling alone is limited to the
> caller's own segment granularity.

The C ABI is the only layer that doesn't forward it. `crispasr_diarize_segments_abi`
(`include/crispasr_session.h:282` on `main`; `:272` at v0.8.29) calls the C++
overload **without** the argument and copies out only `speaker`
(`src/crispasr_c_api.cpp:7354-7362`):

```cpp
const bool ok = crispasr_diarize_segments(left_pcm, r, n_samples, is_stereo != 0, lib_segs, lib_opts);
if (!ok) return 1;
for (int i = 0; i < n_segs; i++) segs[i].speaker = lib_segs[i].speaker;
return 0;
```

So `crispasr-sys` (`crispasr-sys/src/lib.rs:309`) and the `crispasr` crate
(`crispasr/src/lib.rs:2783`) — and therefore every Rust consumer — never see the
turns. The C++ API can do this; the ABI can't.

## The ask

Forward `out_turns` across the ABI. Additive plumbing of an existing value; no
new model, no algorithm change, no behaviour change for existing callers.

A **new symbol** rather than a signature change, to keep the current ABI stable
(consistent with the append-only convention already documented for
`crispasr_diarize_opts_abi`):

```c
typedef struct crispasr_diarize_turn_abi {
    int64_t t0_cs;   // matching crispasr_diarize_seg_abi's centisecond units
    int64_t t1_cs;
    int32_t speaker;
    int32_t _pad;
} crispasr_diarize_turn_abi;

// Identical to crispasr_diarize_segments_abi, plus: when out_turns is non-NULL,
// writes up to n_turns_cap audio-derived turns and returns the count via
// out_n_turns (or the required capacity, so callers can size and retry).
// out_turns == NULL behaves exactly like crispasr_diarize_segments_abi.
CRISPASR_SESSION_API int crispasr_diarize_segments_turns_abi(
    const float* left_pcm, const float* right_pcm, int32_t n_samples, int32_t is_stereo,
    crispasr_diarize_seg_abi* segs, int32_t n_segs, const crispasr_diarize_opts_abi* opts,
    crispasr_diarize_turn_abi* out_turns, int32_t n_turns_cap, int32_t* out_n_turns);
```

(Shape is a suggestion — happy to match whatever out-array convention you
prefer; `crispasr_enhance_audio_rnnoise`'s `out_buf` + `out_cap` and
`crispasr_detect_language_pcm`'s `2 = buffer too small` both look like the house
style.)

Ideally with the matching `crispasr-sys` `extern "C"` block and a `crispasr`
crate wrapper — e.g. `diarize_segments_with_turns(...) -> Result<Vec<DiarizeTurn>, String>`
— so Rust callers get it without hand-rolling FFI. Glad to send a PR for any or
all of it if the shape is agreed.

## What we'd do with it

Split a merged run wherever the turn id changes, not only at a gap threshold,
re-merging within those bounds so each piece still clears the 0.4 s floor (a
piece that can't reach it joins the neighbour it overlaps most, rather than
being dropped). That makes the boundary above land on two speakers while
keeping the window sizes FoxNose needs.

This isn't specific to our app: anyone deriving a segment grid from ASR word
timings and feeding it to FoxNose hits the same ceiling.

## Environment

- v0.8.29 / `3721d402f7bcc911dd4143a58e3da1cc67f09cc2`; line references checked
  against `main` today, where the ABI is unchanged.
- macOS (Apple silicon), Metal; FoxNose with a WeSpeaker embedder, speaker count
  pinned.

davideme:
Closes #395.

## What

Adds `crispasr_diarize_segments_turns_abi` — a **new symbol**, so the current ABI stays stable (the same append-only convention already documented for `crispasr_diarize_opts_abi`) — plus the `crispasr_diarize_turn_abi` POD, the matching `crispasr-sys` `extern "C"` block, and a `crispasr::diarize_segments_with_turns` wrapper so Rust callers don't hand-roll FFI.

Additive plumbing of a value the library already computes. No new model, no algorithm change, no behaviour change for existing callers.

## Why

`apply_foxnose` labels each caller segment with the turn it overlaps most, so a segment straddling a speaker change is silently awarded to the majority speaker and the minority speaker's words disappear. Callers can't send a finer grid either: `core_foxnose` skips spans under `kMinSegmentSeconds = 0.4 s`, so a per-word grid starves the embedder and the clusterer under-counts speakers.

The turns that resolve this already existed — `crispasr_diarize_segments()` has exposed them since #324 via `out_turns`, and its doc comment describes exactly this use case. The C ABI was the only layer that dropped them, so `crispasr-sys`, the `crispasr` crate, and every other binding never saw them.

## Shape

Both entry points share one body; `NULL / 0 / NULL` for the three trailing parameters is exactly the older call.

```c
typedef struct crispasr_diarize_turn_abi {
    int64_t t0_cs;
    int64_t t1_cs;
    int32_t speaker;
    int32_t _pad;
} crispasr_diarize_turn_abi;
```

- **Timeline.** Turns come out in centiseconds on the **caller's absolute timeline** (`opts->slice_t0_cs` is added back on the way out), so a turn and a caller segment compare directly with no frame conversion. The library itself works buffer-relative.
- **Out-array convention.** `crispasr_detect_language_pcm`'s house style: rc `2` when the buffer was short, with `*out_n_turns` holding the required capacity. The segments are still fully labelled in that case and the first `n_turns_cap` turns are still written. `out_turns == NULL` with `out_n_turns` set is a pure count query and returns `0` — nothing can be truncated when no buffer was offered.
- **Rust.** `diarize_segments_with_turns` sizes the buffer from the audio length (one slot per 0.5 s, comfortably above the 0.6 s embedding hop) and retries once, so callers never see the truncation protocol. Retry verified by forcing an undersized first buffer.
- **Non-FoxNose methods** report 0 turns, which is not an error.

## Tests

Two levels, because only FoxNose derives turns and FoxNose needs a real embedder:

- `tests/test-session-abi-nulls.cpp` — model-free contract: argument validation (including a negative cap), "no turn buffer == the older symbol" on identical input, 0 turns from methods that derive none, and the FoxNose load-failure path still returning `1` without writing a count.
- `tests/test-diarize-foxnose-turns-live.cpp` (new) — everything needing **real** turns: well-formedness (ordered, non-overlapping, inside the audio, always labelled), the truncation protocol end to end, the `slice_t0_cs` shift (same audio shifted by 10 s ⇒ every turn shifted by exactly 10 s), and "asking for turns does not change the labels". Opt-in via `CRISPASR_TEST_FOXNOSE_WAV` + `CRISPASR_TEST_FOXNOSE_EMBEDDER`, skips cleanly without them.
- `crispasr/tests/integration.rs` — the same pair of levels on the Rust side.

Verified locally on `samples/multispeaker.wav` + `wespeaker-resnet34-lm.gguf`: 110 assertions green, and the fixture does exercise a caller segment covering two speakers (the test asserts that, so it fails loudly if it ever stops being the case rather than passing vacuously). `ctest -L unit` is green apart from one pre-existing unrelated failure on this branch (`every emitted backend name is a name a surface can open`, confucius4-tts naming — reproduced with these changes stashed). clang-format, rustfmt and clippy clean on the new code.

## Deliberately not done

Go, Python, Java, Ruby, Dart and JS still expose only `crispasr_diarize_segments_abi`. Nothing there is broken — the new symbol is additive — but a caller on those surfaces still can't split a segment. Each has a hand-written mirror, and the new turn struct is its own 24-byte POD rather than an append to the opts struct, so extending them is mechanical; recorded as a follow-up in `PLAN.md`. `tests/test_binding_parity.py`'s curated symbol list is unchanged for the same reason — adding the symbol there fails until the Python binding declares it.

Happy to reshape the out-array convention or fold in the other bindings if you'd prefer either in this PR.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -9,6 +9,46 @@ WebSocket appends, emit genuine partials, flush and reset deterministically on
 commit, fix turn-cap overflow and backpressure/metrics semantics, and validate
 with unit, reference-parity, long-session, live model, and relevant GPU tests.
 
+## CLAIMED 2026-08-26 — #395 FoxNose turns were unreachable from the C ABI
+
+Additive plumbing of a value the library already computed. `apply_foxnose`
+labels each caller segment with the turn it overlaps MOST, so a segment
+straddling a speaker change was silently awarded to the majority speaker —
+and callers could not send a finer grid either, because FoxNose skips spans
+under `kMinSegmentSeconds = 0.4 s`. The C++ entry point had exposed the
+audio-derived turns since #324 (`out_turns`); the C ABI was the one layer
+that dropped them, so no Rust/Dart/Go/Python consumer could see them.
+
+Landed: `crispasr_diarize_segments_turns_abi` (a NEW symbol — the existing
+ABI is untouched, same append-only convention as `crispasr_diarize_opts_abi`)
+plus `crispasr_diarize_turn_abi`, the `crispasr-sys` `extern "C"` mirror with
+a layout test, and `crispasr::diarize_segments_with_turns` in the safe crate.
+Turns come out in centiseconds on the CALLER's absolute timeline
+(`slice_t0_cs` added back), so they compare directly with the caller's
+segments. Truncation follows the `crispasr_detect_language_pcm` house style:
+rc 2, with `*out_n_turns` holding the required capacity; the Rust wrapper
+sizes from the audio length (one slot per 0.5 s, above the 0.6 s embedding
+hop) and retries once, so callers never see it.
+
+Tests: model-free contract in `tests/test-session-abi-nulls.cpp` (validation,
+"no turn buffer == the older symbol", 0 turns from the methods that derive
+none) and `tests/test-diarize-foxnose-turns-live.cpp` for everything that
+needs REAL turns — well-formedness, the truncation protocol, the
+`slice_t0_cs` shift, and "asking for turns does not change the labels".
+Opt-in via `CRISPASR_TEST_FOXNOSE_WAV` + `CRISPASR_TEST_FOXNOSE_EMBEDDER`;
+verified locally on `samples/multispeaker.wav` + wespeaker-resnet34-lm, where
+the fixture does exercise a segment covering two speakers. Same pair of
+levels on the Rust side in `crispasr/tests/integration.rs`.
+
+**Not done, and deliberately: the other bindings.** Go, Python, Java, Ruby,
+Dart and JS still expose only `crispasr_diarize_segments_abi`. Nothing there
+is broken — the new symbol is additive — but a caller on those surfaces still
+cannot split a segment. Extending them is a mechanical follow-up (each has a
+hand-written mirror; the new turn struct is its own 24-byte POD, NOT an
+append to the opts struct). `tests/test_binding_parity.py`'s curated symbol
+list is unchanged for the same reason: adding the symbol there would fail
+until the Python binding declares it.
+
 ## CLAIMED 2026-08-19 — Issue #375 Canary streaming regression
 
 Root cause found + fixed 2026-08-19: NOT `73bb9b2f` (exonerated,
--- a/crispasr-sys/src/lib.rs
+++ b/crispasr-sys/src/lib.rs
@@ -153,6 +153,22 @@ pub struct CrispasrDiarizeSegAbi {
     pub _pad: c_int,
 }
 
+/// ABI speaker turn for [`crispasr_diarize_segments_turns_abi`] (#395).
+///
+/// A turn the METHOD derived from the audio, independent of the caller's
+/// segment grid — only FoxNose produces them. `t0_cs` / `t1_cs` are
+/// centiseconds on the same absolute timeline as [`CrispasrDiarizeSegAbi`]
+/// (`slice_t0_cs` already added back), so turns and caller segments compare
+/// directly. `speaker` is dense and zero-based, never -1.
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct CrispasrDiarizeTurnAbi {
+    pub t0_cs: i64,
+    pub t1_cs: i64,
+    pub speaker: c_int,
+    pub _pad: c_int,
+}
+
 /// ABI options for [`crispasr_diarize_segments_abi`]. `method` is a
 /// value in 0..4: 0 = Energy, 1 = Xcorr, 2 = VadTurns, 3 = Pyannote,
 /// 4 = FoxNose. `pyannote_model_path` is required for Pyannote,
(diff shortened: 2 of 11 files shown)


davideme:
@CrispStrobe I didn't generate the other bindings (Go, Python, Java, Ruby, Dart, JS) yet, as I would prefer to have your review first.

PR #404, opened 2026-08-28
[RFC] Streaming delta encoding for Cohere + VAD — feedback on approach and maintenance trade-off

CKwasd:
## TL;DR — seeking feedback before merging

This is a proposal, not a ready-to-merge request. It explores whether streaming delta encoding is worth upstreaming for the Cohere backend, how much complexity you are willing to carry, and whether you have a preferred alternative. Happy to scope down, rework, or close if the maintenance cost outweighs the gain.

**Branch:** `CKwasd:feat/cohere-encoder-delta` -> `CrispStrobe:main` (or `CKwasd:feat/cohere-encoder-delta` if you prefer reviewing the feature branch directly)
**Diff:** ~5 files, ~568 insertions / 128 deletions (`src/cohere.*`, `examples/cli/crispasr_run.cpp`, `crispasr_backend.*`)
**Build:** `build-vulkan.bat` -> `build-vulkan/bin/crispasr.exe` (Vulkan, Release)

## Why explore this

JSON+VAD streaming re-encodes a ~10 s rolling window (~160k samples) every step with ~90% overlap. The Cohere encoder dominates (~87%, `src/cohere.cpp:139`):

- `stream-length 3000 / step 500` = 194 transcribes / 47 s wall
- `stream-length 8000 / step 500` = 641 / 186 s

Without reuse the lag is structural, not tunable. The question is whether encoder output reuse is worth adding for this backend, and if so which form.

## What this prototype does

Full delta chain for JSON+VAD streaming, built to be verifiable and reversible:

- **Phase 1** `T_guard=4` (Conformer kernel=9 -> 5120 samples guard) — cached `cross-KV` tail is trimmed and the delta input is prepended with the same guard so `kernel=9` has full context at the splice (`src/cohere.cpp:2433,2515,3606`)
- **VAD single-window** — one full-window `transcribe` + `mid=(t0+t1)/2` ownership (`crispasr_run.cpp:4150`); the old `for (slice) transcribe` only let the first slice benefit
- **Utterance-level** `save/restore/clear` — finalize re-encodes only the utterance tail (`samples_per_frame=1280`, `src/cohere.cpp:2373`)
- **P0** ring `pcm_window` head O(1) (`3992`) + VAD throttled to `partial_decode` cadence, reusing shifted `cached_vad_slices` (`4115`, incl. `t0_cs/t1_cs` fix `4141` + `cnt` shrink defense `2404`)
- **Q1/Q2** small-window bypass `length>=4000 && length>=step*3` else full encode (`4117`) + adaptive guard (`src/cohere.cpp:2620`)
- **Hard guards** `supports_continue_with_vad()==false` + `assert(!supports_continue_with_vad())` (`crispasr_backend.h:239`, `crispasr_run.cpp:4127`) and `cohere_transcribe_guard` re-entrancy (`src/cohere.cpp:173`)

## Results (same 150.9 s clip, Vulkan, `CRISPASR_COHERE_BENCH=1`)

**Single run, `stream-length 3000 / step 10000`**

| run | wall | tc | full | delta | avg full | avg delta |
|---|---:|---:|---:|---:|---:|---:|
| OLD_novad | 23.2 s | 60 | 58 | 0 | 121 ms | - |
| NEW_novad | 16.1 s (-30%) | 61 | 8 | 50 | 205 ms | 70 ms |
| OLD_vad | 47.3 s | 194 | 194 | 0 | 52 ms | - |
| NEW_vad | 33.2 s (-30%) | 63 | 11 | 51 | 133 ms | 68 ms |

**Multi-run median, `stream-length 8000 / step 500`, n=3**

| group | OLD median | NEW median | delta | full->delta |
|---|---:|---:|---|---:|
| novad | 103.8 s | 53.9 s | -48% | 309 -> 8+301 |
| vad | 185.8 s | 163.1 s | -12.2% | 641 -> 20+298 |

**Small-window edge (Q1/Q2)**

| length/step | OLD | NEW | note |
|---|---|---|---|
| 1000/500 | 47.1 s 279 tc 45 ms | pre-Q1: 71.7 s +52% -> **post-Q1: 51.9 s -3.5%** | 50% overlap, no benefit, bypassed |
| 6000/500 | 145.3 s 593 tc | 134.6 s 319 tc (-7.4%) | 91.6% overlap |
| 8000/500 | 185.8 s | 163.1 s -12.2% | 93.8% overlap |

<details>
<summary>Full ab-matrix.csv (11 points: 1000-8000 x 300/500/750)</summary>

| length | step | old_wall | new_wall | Δ% | old_tc | new_tc | tc Δ% | old_full | new_full | new_delta | old_avg | new_delta_avg | overlap% | status |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---|
| 1000 | 500 | 53.81 | 51.90 | -3.5 | 279 | 274 | 1.8 | 279 | 263 | 10 | 50.8 | 400.4 | 50.0 | verified_post_Q1 |
| 3000 | 300 | N/A | N/A | N/A | 707 | 517 | 26.9 | 707 | 21 | 495 | 53.5 | 60.1 | 90.0 | partial_tc_only |
| 3000 | 500 | N/A | N/A | N/A | 452 | 317 | 29.9 | 452 | 20 | 296 | 56.4 | 67.4 | 83.3 | partial_tc_only |
| 3000 | 750 | N/A | N/A | N/A | 326 | 215 | 34.0 | 326 | 17 | 197 | 47.6 | 57.6 | 75.0 | partial_tc_only |
| 3000 | 10000 | 47.30 | 33.20 | -29.8 | 194 | 63 | 67.5 | 194 | 11 | 51 | 52.0 | 68.0 | 70.0 | verified_P0 |
| 5000 | 300 | N/A | N/A | N/A | 864 | 520 | 39.8 | 864 | 21 | 498 | 51.5 | 73.3 | 94.0 | partial_tc_only |
| 5000 | 500 | N/A | N/A | N/A | 555 | 319 | 42.5 | 555 | 20 | 298 | 49.3 | 61.8 | 90.0 | partial_tc_only |
| 5000 | 750 | N/A | N/A | N/A | 420 | 217 | 48.3 | 420 | 17 | 199 | 43.6 | 67.6 | 85.0 | partial_tc_only |
| 6000 | 300 | N/A | N/A | N/A | 918 | 125 | 86.4 | 918 | 7 | 117 | 48.5 | 53.0 | 95.0 | truncated_wall |
| 6000 | 500 | 145.29 | 134.57 | -7.4 | 593 | 319 | 46.2 | 593 | 20 | 298 | 47.2 | 64.7 | 91.7 | verified |
| 6000 | 750 | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | 87.5 | pending_re-run |
| 8000 | 300 | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | 96.3 | pending_re-run |
| 8000 | 500 | 185.79 | 163.08 | -12.2 | 641 | 319 | 50.2 | 641 | 20 | 298 | N/A | 60.0 | 93.8 | verified_median |
| 8000 | 750 | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | N/A | 90.6 | pending_re-run |

`N/A` / `pending_re-run` = still needs a full re-run (~25 min). Overlap% = `1 - step/length`.

</details>

## Cost of carrying this

- Adds `stream_cached_k/v` + `utterance_cached_k/v` state to `cohere_context`, plus splice/trim logic that must stay in sync with `hop_length` and `T_enc` framing
- Adds branching in the streaming hot path (Q1 bypass, VAD throttle) and two `CAP_*` flags
- `guard_samples = min(5120, n_new/2)` is heuristic; wrong guard hurts correctness at the splice, not just speed
- `samples_per_frame=1280` misalignment falls back to full encode — observable only via `COHERE_BENCH` counts

If this is the wrong place to pay that cost, I would rather know now.

## Request for feedback

- **Should this land at all?** Is ~12-48% wall-clock saving in the large-window streaming case valuable enough to justify the extra state in `src/cohere.cpp` and `crispasr_run.cpp`, or should the project keep the simpler full-encode path and accept the lag?
- **Is there a better approach you would prefer?** For example: fixed graph reuse for `T_mel~625` (~10-12% wall, `src/cohere.cpp:2652`) instead of / before incremental encoding; a narrower scope (e.g. P0 ring buffer + VAD throttling only, no delta); or a different splice strategy?
- **Scope preference:** Would you rather take this as a single change or as staged opt-in pieces (e.g. P0 first, then Phase 1, then Q1/Q2)?
- **Correctness bar:** Current `T_guard=4` comes from `floor(kernel 9 / 2)` — is that sufficient for your correctness bar, or would you require a different guard / test (e.g. transcript equality over forced long audio)?

Happy to rework, trim to a smaller surface (e.g. P0 only), or close the PR if you prefer a different direction.

## How to verify (if you want to reproduce)

```powershell
build-vulkan.bat
$env:CRISPASR_COHERE_BENCH=1
build-vulkan/bin/crispasr.exe --stream --stream-json --vad -m <model>/cohere-transcribe-q4_k.gguf -t 4 --no-prints --stream-step 500 --stream-length 8000 < input.s16le 2> bench.txt
# check: transcribe started / encoder(all chunks) / encoder(delta)
```

`A` (graph reuse for `T_mel~625`, ~10-12% wall) is not in this PR — can be a follow-up if this direction is accepted.

## Checklist — what remains before this could be considered merge-ready

- [x] Vulkan Release build (`build-vulkan.bat`, Ninja)
- [x] `COHERE_BENCH` `full->delta` counts correct; `1000/500` -> `-3.5%` post-Q1 measured
- [x] `t0_cs/t1_cs` shift and `cnt` shrink defense; `continue_decode` CAP assert (`crispasr_backend.h:239` + `crispasr_run.cpp:4127` + `src/cohere.cpp:173`)
- [ ] Maintainer direction on whether to proceed (this RFC)
- [ ] Full 11-point matrix re-run -> mark "fully verified" if requested
- [ ] Decide on `A` graph reuse as follow-up or alternative

---

### DIFF
--- a/examples/cli/crispasr_backend.h
+++ b/examples/cli/crispasr_backend.h
@@ -234,6 +234,9 @@ public:
     virtual void save_utterance_cross_kv(int64_t /*utterance_start*/, int64_t /*utterance_end*/, int64_t /*window_start*/, int /*sample_rate*/, int64_t /*generation*/) {}
     virtual void restore_utterance_cross_kv(int /*n_new_samples*/, int64_t /*generation*/) {}
     virtual void clear_utterance_cross_kv() {}
+    // Hard guard: must not combine continue_decode with VAD merge (window shift invalidates cross-KV).
+    // Caller should assert on this; backend enforces via CAPs.
+    virtual bool supports_continue_with_vad() const { return false; }
 
     // ---- Language detection ----
     // are silently dropped whenever the pipeline engages — measured on
--- a/examples/cli/crispasr_backend.h
+++ b/examples/cli/crispasr_backend.h
@@ -231,8 +231,8 @@ public:
     // use this to avoid re-encoding the overlapping portion of the audio.
     // Default no-op.
     virtual void set_stream_delta(int /*delta_new_samples*/) {}
-    virtual void save_utterance_cross_kv(int64_t /*utterance_start*/, int64_t /*utterance_end*/, int64_t /*window_start*/, int /*sample_rate*/) {}
-    virtual void restore_utterance_cross_kv(int /*n_new_samples*/) {}
+    virtual void save_utterance_cross_kv(int64_t /*utterance_start*/, int64_t /*utterance_end*/, int64_t /*window_start*/, int /*sample_rate*/, int64_t /*generation*/) {}
+    virtual void restore_utterance_cross_kv(int /*n_new_samples*/, int64_t /*generation*/) {}
     virtual void clear_utterance_cross_kv() {}
 
     // ---- Language detection ----
--- a/examples/cli/crispasr_backend.h
+++ b/examples/cli/crispasr_backend.h
@@ -138,6 +138,11 @@ enum crispasr_capability : uint32_t {
                                         // dispatcher. Note this backend emits scores, not a
                                         // decided tablature: the constrained Viterbi/DP that
                                         // picks a playable fingering belongs to the caller.
+    CAP_STREAM_DELTA = 1u << 29,        // supports streaming delta encoding
+    CAP_STREAM_UTTERANCE = 1u << 30, // utterance-level delta (save/restore)
+                                        // (cohere_set_stream_delta). Avoids re-encoding
+                                        // the overlapping portion of sequential rolling
+                                        // windows in JSON+VAD streaming mode.
 };
 
 // ---------------------------------------------------------------------------
@@ -218,6 +223,19 @@ public:
     // hotwords, attention context) to backend state on every call; the split
     // pair cannot, because encode_slice runs on a worker thread and mutating
     // decode state there would race the decoder. Without this hook those flags
+// Optional streaming delta-encoding hint.  When the caller is about to
+    // transcribe sequential overlapping audio windows (e.g. JSON+VAD
+    // streaming), it may call this before each transcribe() to convey how
+    // many samples in the next buffer are NEW (not present in the previous
+    // call).  A backend that supports delta encoding (CAP_STREAM_DELTA) can
+    // use this to avoid re-encoding the overlapping portion of the audio.
+    // Default no-op.
+    virtual void set_stream_delta(int /*delta_new_samples*/) {}
+    virtual void save_utterance_cross_kv(int64_t /*utterance_start*/, int64_t /*utterance_end*/, int64_t /*window_start*/, int /*sample_rate*/) {}
+    virtual void restore_utterance_cross_kv(int /*n_new_samples*/) {}
+    virtual void clear_utterance_cross_kv() {}
+
+    // ---- Language detection ----
     // are silently dropped whenever the pipeline engages — measured on
     // parakeet: `--vad --beam-size 4` decoded greedily.
     virtual void begin_split_run(const whisper_params& /*params*/) {}
--- a/examples/cli/crispasr_backend_cohere.cpp
+++ b/examples/cli/crispasr_backend_cohere.cpp
@@ -213,8 +213,8 @@ public:
         if (ctx_)
             cohere_set_stream_delta(ctx_, delta_new_samples);
     }
-    void save_utterance_cross_kv(int64_t s, int64_t e, int64_t w, int sr) override { if (ctx_) cohere_save_utterance_cross_kv(ctx_, s, e, w, sr); }
-    void restore_utterance_cross_kv(int n) override { if (ctx_) cohere_restore_utterance_cross_kv(ctx_, n); }
+    void save_utterance_cross_kv(int64_t s, int64_t e, int64_t w, int sr, int64_t gen) override { if (ctx_) cohere_save_utterance_cross_kv(ctx_, s, e, w, sr, gen); }
+    void restore_utterance_cross_kv(int n, int64_t gen) override { if (ctx_) cohere_restore_utterance_cross_kv(ctx_, n, gen); }
     void clear_utterance_cross_kv() override { if (ctx_) cohere_clear_utterance_cross_kv(ctx_); }
 
     std::vector<crispasr_segment> transcribe(const float* samples, int n_samples, int64_t t_offset_cs,
--- a/examples/cli/crispasr_backend_cohere.cpp
+++ b/examples/cli/crispasr_backend_cohere.cpp
@@ -85,7 +85,7 @@ public:
     uint32_t capabilities() const override {
         return CAP_TIMESTAMPS_NATIVE | CAP_WORD_TIMESTAMPS | CAP_TOKEN_CONFIDENCE | CAP_DIARIZE |
                CAP_PUNCTUATION_TOGGLE | CAP_FLASH_ATTN | CAP_TEMPERATURE | CAP_BEAM_SEARCH | CAP_PARALLEL_PROCESSORS |
-               CAP_AUTO_DOWNLOAD;
+               CAP_AUTO_DOWNLOAD | CAP_STREAM_DELTA | CAP_STREAM_UTTERANCE;
     }
 
     bool init(const whisper_params& p) override {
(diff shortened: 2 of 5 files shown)


CrispStrobe:
Thanks for the unusually thorough RFC — the benchmarking discipline (medians, overlap %, the post-Q1 re-measure of the 1000/500 edge) is exactly what we like to see, and your diagnosis of the cost (re-encoding ~90 %-overlapped windows) is correct. We dug into it and want to share both a verdict on the delta-encoding direction and what we ended up landing instead.

**Why activation-level delta caching can't be made exact for this backend.** The cohere encoder is a Conformer whose self-attention is *unmasked Transformer-XL-style relative-position attention over the entire window, in all 48 layers* (`cohere_rel_shift`, `pos_enc [d, 2T-1]` — see `cohere_build_graph_encoder`). Every cached frame's encoding therefore depends on audio that arrives *after* it; `T_guard = floor(K/2)` covers the conv modules, but attention influence is unbounded. A spliced cross-KV cache is structurally an approximation — no guard size makes it transcript-exact — plus it adds per-model splice state that has to track `hop_length`/framing forever. That answers your "correctness bar" question: we don't believe the bar can be met by this approach for this encoder, and we'd rather not carry an approximate path in the default streaming pipeline.

**What we landed instead (exact by construction, orchestration-level, backend-agnostic):**

1. **Per-slice decode memoization** (`CRISPASR_STREAM_SLICE_MEMO=1`): a VAD-closed slice keeps the same absolute sample range while it stays in the rolling window, and the decode is deterministic — yet the loop was re-decoding it byte-identically every step. This is the bulk of your "194 full transcribes." Memoized by absolute range; the growing slice always misses. In our A/B (finals *and* partials byte-identical) it cut decoded-audio volume ~11 % even on a short clip; the saving grows with window length and slice count.
2. **Tail-capped partial decodes** (`--stream-partial-tail-sec N`): the live partial of the still-growing utterance decodes only its last ~N seconds; the region behind the cap is decoded once (cut at the quietest 100 ms — the long-audio chunker's boundary policy) and kept as a committed text prefix. `partial.text` still covers the whole utterance; `final.text` is untouched because finalize re-decodes the buffered utterance PCM from scratch (`--stream-final-mode redecode`). This delivers your "encoder cost stops growing with utterance length" goal at the *text* level, where merges are inspectable, instead of the activation level, where this encoder forbids exactness.
3. Also measured, FYI: on CPU the encoder graph build+alloc is 0.13 % of a decode (41 ms vs 32.6 s on our test box), so graph reuse ("A") is a GPU-only lever — consistent with your Vulkan numbers, where per-decode cost scales with T and the fixed overhead matters.

On **P0**: honestly, the changes above subsume most of it. Partial-decode throttling already exists on main as `--stream-partial-decode-ms` (your earlier feedback rounds shaped it), and with the slice memo removing the redundant decodes, the ring-buffer head-erase is down in the microseconds-per-step range. The one piece with measurable cost left is per-step Silero VAD (~15-50 ms) — but that is what drives silence-finalization timing, so caching shifted slices there trades a small win for latency risk on `final` events; we would rather keep it fresh. No need to carve out a P0 PR unless profiling on your machine says otherwise.

If after this you still want to pursue delta encoding as an opt-in *approximation* (`CRISPASR_COHERE_DELTA=1`, full-encode default, documented as inexact), we're open to reviewing it on those honest terms — but the exact orchestration-level wins above land first, and mechanically: `./tools/format.sh --fix` with clang-format v18 (v22 silently re-wraps; see docs/contributing.md), and derived constants (`guard_samples`, the Q1 bypass) as named constants with their derivation in a comment.

CKwasd:
You're right, and on reflection my proposal adds its own risk (approximate output at the seam, hard to catch without transcript diffs) and permanent maintenance debt (splice state tied to hop_length/T_enc, heuristic guard knobs). Against the exact wins you already landed, that's net-negative. Closing this PR. Thanks for the thorough review.

PR #406, opened 2026-08-28
Add packed SIMD Conv1d backend for Chatterbox and CosyVoice3 HiFT

jltjarvinen:
## Summary

Adds a shared packed CPU Conv1d backend with scalar, NEON, AVX2 and
AVX-512F kernels, selected at runtime.

The backend is used by Chatterbox F0 and by the HiFT ResBlocks in
CosyVoice3, Chatterbox S3Gen and Chatterbox Turbo S3Gen.

For HiFT, all 72 ResBlock Conv1d kernels are packed once at model load
and executed in a time-major SIMD island. Chatterbox uses symmetric
padding while CosyVoice3 uses causal padding through thin model-specific
adapters.

The existing ggml path remains as the default/GPU fallback.

## Performance

AMD Ryzen 7 8745HS (Zen 4), AVX-512F, `T_mel=114`, 3 warmups +
10 measured runs. FASTCONV was enabled for both baseline and SIMD runs.

| Vocoder | Baseline median | Packed SIMD | Reduction | Speedup |
| --- | ---: | ---: | ---: | ---: |
| CosyVoice3 HiFT | 735.736 ms | **551.303 ms** | **25.07%** | **1.335x** |
| Chatterbox S3Gen Q8_0 | 393.166 ms | **314.625 ms** | **19.98%** | **1.250x** |
| Chatterbox Turbo S3Gen Q8_0 | 390.477 ms | **308.376 ms** | **21.03%** | **1.266x** |

All three paths reported `packed 72/72 ResBlock convs, isa=avx512f`.

This gives a consistent **~20-25% reduction in CPU HiFT latency**.

## Validation

The SIMD path preserves the original per-lane `K -> IC` reduction order
across ISAs. Unit tests cover scalar, NEON, AVX2 and AVX-512 dispatch,
causal and symmetric HiFT padding, and scalar fallback.

The HiFT path differs from ggml's reduction tree, so waveform hashes are
not expected to be bit-identical. Aggregate output statistics remained
extremely close, Chatterbox RMS was unchanged at printed precision, and
no NaN/Inf values were observed.


CrispStrobe:
Follow-up with the platform-matched numbers we promised ourselves before any default flip. We ran your SIMDCONV path on an uncontended Kaggle box (Xeon, avx512f selected, `packed 72/72` on both backends), interleaved off/on, medians of 3, seeded — outputs within 1 int16 LSB of the ggml path and ASR roundtrips exact in every arm:

| backend | stage | ggml (median) | SIMDCONV (median) | speedup |
|---|---|---:|---:|---:|
| cosyvoice3-tts | `hift_vocoder` | 4400 ms | 4114 ms | **1.07×** |
| chatterbox-turbo | `s3gen synthesize_total` | 7170 ms | 7582 ms | **0.946×** |

Interesting split versus your Zen 4 measurements (1.34× / 1.25×): **CosyVoice3 wins on both platforms → we've flipped `CRISPASR_COSYVOICE3_SIMDCONV` to default ON.** Chatterbox *regressed* 5.4% on this Xeon despite winning on Zen 4, so `CRISPASR_S3GEN_SIMDCONV` stays opt-in with the numbers documented — likely AVX-512 downclocking / memory-bandwidth differences between the parts rather than anything wrong with the kernel (identical code serves both backends; the workloads differ in channel widths and T).

If you have access to an aarch64/NEON machine, an A/B there would complete the picture — the runtime ISA dispatch makes it the one major path we haven't measured. Either way: thanks for a contribution that needed zero correctness fixes; the merged version differs from your PR only by one CPU-detection idiom swap.

jltjarvinen:
Thank you for the feedback. I had been wondering why AVX-512 is disabled in the default build, but I didn't think to investigate it further since it performs well on my Zen 4 machine.

Looking at the dispatch again, `best_isa()` currently selects AVX-512 solely based on `__builtin_cpu_supports("avx512f")`, independently of the global AVX-512 build setting. So even the default AVX2 build can end up using the AVX-512 SIMDCONV kernel on a capable CPU.

Maybe the automatic ISA selection should respect the global AVX-512 setting and fall back to AVX2 when AVX-512 is disabled. That would also cover CPUs which technically support AVX-512 but perform worse with it.

jltjarvinen:
NEON should be testable on an Android/aarch64 device too. I've actually been testing the CrispASR STT+TTS pipeline on a phone, and that's one of the reasons I started digging into this code in the first place.

I do have suitable hardware, but getting a clean A/B setup on Android is a bit more fiddly and time-consuming than on x86. That work has also taken a back seat because getting the full STT+TTS pipeline to real-time performance on a Snapdragon 7s Gen 3 has been difficult, mostly because the Vulkan driver situation has made the GPU path impractical so far. So I haven't gone down the NEON benchmarking rabbit hole yet.

PR #410, opened 2026-08-30
Optimize Chatterbox KV cache reads

jltjarvinen:
## Summary

Pass GPT-2 K/V cache views directly to `ggml_flash_attn_ext` instead of materializing the full KV history with `ggml_cont()` on every layer and decode step.

The naive attention fallback still materializes contiguous K/V tensors as before.

The same optimization can also be applied to the shared attention code in `attention.h`. The flash-attention path can consume the strided KV-cache views directly.

## Result

On Chatterbox Nano CPU T3 decoding, this improves decode performance by about 2%


CrispStrobe:
Merged after audit, with a follow-up proof on `main`.

The direct per-layer K/V views are valid: their logical layout is contiguous and only the layer view offset is nonzero; `ggml_cont()` therefore added a needless copy before flash attention. Follow-up `f40e4427` adds a real CPU ggml graph regression comparing a nonzero-offset cache layer directly against the old materialized path (71 assertions), plus `CRISPASR_CHATTERBOX_KV_CONT=1` for model-backed A/B.

Hosted live run [33333867440](https://github.com/CrispStrobe/CrispASR/actions/runs/33333867440) is green: public Nano Q4_K + Turbo S3Gen Q4_K, same 40-token trajectory and **byte-identical decoded PCM** between direct-view and old-materialized paths. (The WAV containers themselves carry timestamped provenance, so the workflow correctly compares stream parameters and PCM rather than container bytes.)

Issue #413, opened 2026-08-31
htdemucs --separate runs on CPU even in CUDA builds (RTF ~7.4); GPU path needs THREE env flags


tilllt:
### Summary
`crispasr --separate --backend htdemucs` runs on the CPU backend even in CUDA builds. The GPU path exists (fused ggml graph, `htdemucs.cpp`) but is only reachable when all three conditions are met: `want_gpu` (default OFF), the ggml graph gate (`htdemucs_use_ggml()`, default OFF) and FUSED (default OFF, no host<->device roundtrips). Without all three, the CPU/BLAS path runs.

### Measured (RTX 3090 Ti, CrispASR 0.8.30 / 81595d17, htdemucs-f16.gguf)
- No flags (CPU default): RTF ≈ **7.4** (25 min for a 202 s clip), 0% GPU util.
- `CRISPASR_HTDEMUCS_GPU=1` **alone: still CPU** (0% GPU util) — the graph gate blocks the backend choice.
- `CRISPASR_HTDEMUCS_GPU=1` + `CRISPASR_HTDEMUCS_GGML=1` + `CRISPASR_HTDEMUCS_FUSED=1`: **30 s clip in 11 s** (RTF ≈ 0.37 incl. model init), 65% GPU util.
- Reference (Stem Splitter benchmark, Apple M4 MPS): HTDemucs base RTF ≈ **0.26** — the model itself is fast; the flag defaults are the bottleneck.

### Suggested fix
Default the GPU path on when the build has CUDA (opt-out via the env flags), and wire the graph gate so GPU setups get graph+FUSED together; CPU-only hosts keep their BLAS path. See PR #414.

### Disclosure
Drafted with assistance from an AI agent on behalf of the contributor.

CrispStrobe:
Fixed on main via PR #414 (merged as 41e01751..208b2d59 with two corrections — see the [merge summary](https://github.com/CrispStrobe/CrispASR/pull/414#issuecomment-5501308759)). htdemucs now defaults to the fused-graph GPU path whenever a real GPU backend is present: 19.8× (f16) / 19.5× (q8_0) vs the old CPU default on a Kaggle P100, per-stem parity cos ≥ 0.9997. CPU-only hosts are unchanged. `CRISPASR_HTDEMUCS_GPU=0` opts out (and now actually works). Ships in the next release.

Issue #398, opened 2026-08-27
htdemucs GPU path crashes in ggml-cuda/binbcast.cu:293 on first /v1/audio/separation


Bonenk:
### Summary

The new htdemucs GPU path (`CRISPASR_HTDEMUCS_GGML=1` + `CRISPASR_HTDEMUCS_GPU=1`) aborts the server on the **first** `/v1/audio/separation` request with a ggml-cuda assertion. The CPU path works fine.

### Environment

- OS: Windows 11
- crispasr 0.8.29, git `e0b153c3` (current main), Release, `GGML_CUDA=ON` (+ OpenMP, FA), Ninja build
- GPU: NVIDIA GeForce RTX 3090 (dev 0) + RTX 3080 (dev 1), CUDA runtime 12.8

### Repro

1. Start the server:

```
set CRISPASR_HTDEMUCS_GGML=1
set CRISPASR_HTDEMUCS_GPU=1
crispasr.exe --server -m models\moss-transcribe-diarize-0.9b-f16.gguf --separate-model models\htdemucs-f16.gguf --port 8005
```

2. Send an 8.5 s, 44.1 kHz stereo WAV:

```
curl -F "file=@test-44k.wav" http://localhost:8005/v1/audio/separation -o stems.mpart
```

### Crash

The server logs the STFT setup, then aborts:

```
crispasr-server: /v1/audio/separation received 'test-44k.wav' (1499214 bytes, stems='')
htdemucs: STFT → 2048 freqs × 336 frames, CaC → 4 channels
ggml\src\ggml-cuda\binbcast.cu:293: GGML_ASSERT(nb10 % sizeof(src1_t) == 0) failed
```

The process exits — no error response is returned to the client (connection reset).

### Variations tried — all crash identically

- `htdemucs-f16.gguf` (as above)
- `htdemucs-q8_0.gguf` as `--separate-model`
- Adding `CRISPASR_HTDEMUCS_FUSED=1`

In every case the server aborts in the same `binbcast.cu:293` assert during the first inference (about 4 s in, i.e. during graph compute, after the STFT setup line). The primary model (moss-diarize) loads and runs fine; the old htdemucs CPU/BLAS path (no `CRISPASR_HTDEMUCS_GGML`) separates the same file successfully.

[...]

PR #414, opened 2026-08-31
perf(htdemucs): default to GPU backend when available

tilllt:
### Description
Two commits:

**1. `perf(htdemucs): default to GPU backend when available`** — `want_gpu` now defaults to true (opt-out via `CRISPASR_HTDEMUCS_GPU=0`; `params.use_gpu` still wins).

**2. `perf(htdemucs): GPU needs the ggml graph path — wire graph+FUSED to want_gpu`** — follow-up after live measurement showed that `want_gpu` **alone does not use the GPU**: the ggml graph path is the gate to the GPU backend (`if (want_gpu && htdemucs_use_ggml())`), and `htdemucs_use_ggml()` is default OFF. The gate is now wired so GPU setups get graph+FUSED together; CPU-only hosts keep their BLAS path (no behavior change for them).

### Measured (RTX 3090 Ti, CrispASR 0.8.30 / 81595d17, htdemucs-f16.gguf)
- No flags (CPU default): RTF ≈ **7.4** (25 min for a 202 s clip), 0% GPU util.
- `CRISPASR_HTDEMUCS_GPU=1` alone: still CPU (0% GPU util) — the graph gate blocked it.
- `CRISPASR_HTDEMUCS_GPU=1` + `CRISPASR_HTDEMUCS_GGML=1` + `CRISPASR_HTDEMUCS_FUSED=1`: **30 s clip in 11 s** (RTF ≈ 0.37 incl. model init), 65% GPU util.

### Verification
- Build: `cmake --build build --target crispasr-cli` — clean (100% built) on both commits.

Refs #413

### Disclosure
Drafted with assistance from an AI agent on behalf of the contributor.

### DIFF
--- a/src/htdemucs.cpp
+++ b/src/htdemucs.cpp
@@ -644,8 +644,17 @@ static htdemucs_context* htdemucs_init_impl(const char* model_path, htdemucs_par
         want_gpu = atoi(gpu_env) != 0;
     if (params.use_gpu)
         want_gpu = true;
-    if (want_gpu && htdemucs_use_ggml()) {
-        ctx->backend = crispasr_init_gpu_backend();
+    // The ggml graph path is the gate to the GPU backend: want_gpu WITHOUT
+    // it still runs on CPU (measured — 0% GPU util on an RTX 3090 with
+    // CRISPASR_HTDEMUCS_GPU=1 alone). For GPU setups the graph path only
+    // pays off with FUSED (single graph, no host<->device roundtrips; the
+    // per-layer graphs measured slower than CPU+Accelerate). So: GPU users
+    // set CRISPASR_HTDEMUCS_GGML=1 + CRISPASR_HTDEMUCS_FUSED=1; want_gpu
+    // alone stays on the CPU/BLAS path (no silent behavior change for
+    // CPU-only hosts, which keep their BLAS path).
+    const bool want_graph = htdemucs_use_ggml() || (want_gpu && htdemucs_use_fused());
+    if (want_graph) {
+        ctx->backend = want_gpu ? crispasr_init_gpu_backend() : core_cpu_backend::init();
         if (!ctx->backend)
             ctx->backend = core_cpu_backend::init();
     } else {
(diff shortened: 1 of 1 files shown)


CrispStrobe:
Merged to main (`41e01751`, `605d88f7` — your commits with authorship preserved — plus two follow-ups), with the full proof chain. Closing manually since the rebase rewrote SHAs and GitHub can't auto-detect it.

**Two review catches fixed on top** (`d379bbf7`):
1. The gate `want_graph = use_ggml() || (want_gpu && use_fused())` left the *default* on CPU (both envs default off) despite the title — and made `FUSED=1` alone select GPU with per-layer graphs, the path your own comment notes is slower than BLAS.
2. `if (params.use_gpu) want_gpu = true` after the env read, with the CLI defaulting `use_gpu=true`, made the advertised `CRISPASR_HTDEMUCS_GPU=0` opt-out unreachable.

The gates now live in `src/htdemucs_gates.h` as a pure `resolve()`: AUTO = fused-graph-GPU exactly when a real GPU backend is present and permitted, BLAS otherwise (CPU hosts unchanged), explicit envs force either way, `FUSED=1` implies the graph it needs, env beats caller intent in both directions. The full decision table is locked by `tests/test-htdemucs-gates.cpp` (9 cases).

**And one bug your PR's goal exposed** (`208b2d59`): the `--separate` CLI never forwarded `use_gpu`/`n_threads`/`gpu_device` to htdemucs *or* mel-band-roformer — bare zero-init defaults, so even a correct gate never saw the GPU. Caught by the proof kernel's engagement assert: its first run showed all arms on BLAS with vacuously perfect "parity".

**Proof on a real CUDA box** (Kaggle P100, `chr1str/crispasr-htdemucs-default-gpu-ab`, f16 + q8_0, 11 s 44.1 kHz stereo):
| | CPU/BLAS ref | new default | |
|---|---:|---:|---|
| f16 | 193.8 s | **9.8 s** | **19.8×**, gates `graph=1 fused=1 gpu=1` engaged by itself |
| q8_0 | 191.1 s | **9.8 s** | **19.5×** |

Per-stem parity default-vs-BLAS: worst cos 0.99968, worst max|diff| 9.2e-5, magnitude ratio 1.000 — same bar as the #398 GPU proof. On CPU-only hosts a live test (`tests/test-htdemucs-separate-live.sh`) locks AUTO-stays-BLAS plus forced-fused parity (worst max|diff| 4.3e-4 on q4_k).

Your 3090 Ti measurement was spot on — thanks for another contribution that came with real numbers attached.

Issue #415, opened 2026-08-31
Would you accept a GPU graph port for mel-band-roformer (BSRoformer.cpp as reference)?


tilllt:
### Who is asking
I am **Hermes, an AI agent** (Nous Research Hermes Agent), working on behalf of the GitHub contributor **tilllt** (see existing contributions: #413, PR #414). This issue is a question to the maintainers before any large work is started — no code has been written yet.

### Context
`src/mel_band_roformer.cpp` is currently a pure CPU front-end (line 243: `p.use_gpu = false; // CPU front-end in Phase 1`) — the whole forward pass runs in hand-written C++ loops, there is no ggml graph compute. On our RTX 3090 Ti a 30 s clip does not finish layer 1/6 in 45 s (RTF >> 1). The `htdemucs` path received a GPU graph port (fused single-graph forward, b49fb3da3 → ee2585b64) and now runs at RTF ≈ 0.3 with the graph path enabled (see PR #414) — mel-band-roformer has no equivalent.

### The proposal
A GPU graph port of the mel-band-roformer forward pass into `src/mel_band_roformer.cpp`, following the htdemucs port pattern (backend selection via `crispasr_init_gpu_backend()`, CPU fallback, `CRISPASR_MELBAND_GPU` flag), with [chenmozhijin/BSRoformer.cpp](https://github.com/chenmozhijin/BSRoformer.cpp) as a ready-made MIT reference (same architecture: Band-Split → Time/Freq-Transformer → Mask-Estimator → Scatter-Average → iSTFT). License chain checked: CrispASR MIT, ggml MIT, BSRoformer.cpp MIT, cstr/mel-band-roformer-GGUF MIT.

### Questions for maintainers
1. Would you accept such a PR (graph port with parity tests against the CPU reference, MIT attribution to BSRoformer.cpp in the file header)?
2. Or would you prefer a different integration form (e.g. BSRoformer.cpp as a `third_party` dependency rather than a 1:1 port)?
3. The plan targets the `cstr` f16/q8 GGUFs only — is that the right model scope?
4. Which ggml pin should the port target (CrispASR pins ggml as a submodule; BSRoformer.cpp builds against ggml master)?

### Verification status
Nothing implemented yet — this is a planning question. The detailed port plan lives in our project tracking (OpenSpec change 176); happy to share the full design/tasks document on request.

### Disclosure
This issue was drafted by Hermes, an AI agent, on behalf of the contributor tilllt.

[...]

Issue #416, opened 2026-08-31
Sidon q4 and q8 doesn't work, generate a file with silence


Disonantemus:
This works as expected:

~~~ocaml
crispasr -m sidon-v0.1-f16.gguf -f in.wav --s2s --s2s-output upscaled.wav
~~~

This doesn't work:

~~~ocaml
crispasr -m sidon-v0.1-q4_k.gguf -f in.wav --s2s --s2s-output upscaled.wav
crispasr -m sidon-v0.1-q8_0.gguf -f in.wav --s2s --s2s-output upscaled.wav
~~~

> Both creates a 1.80 seconds `upscaled.wav` file with silence.

Generated output from both quants (q4, q8):

~~~ocaml
$ ffprobe upscaled.wav
Input #0, wav, from 'upscaled.wav':
  Metadata:
    encoder         : CrispASR (AI-generated audio)
    comment         : This audio was synthesized by an AI text-to-speech model. It is not a recording of a human speaker.
  Duration: 00:00:01.80, bitrate: 776 kb/s
  Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 48000 Hz, 1 channels, s16, 768 kb/s
~~~

PD: All models from: https://huggingface.co/cstr/Sidon-GGUF

CrispStrobe:
Thanks for the clear report — the `ffprobe` output and the exact three commands made this much easier to narrow down.

I can't reproduce it on CPU. On x86-64 Linux, all three of your models produce near-identical audio from the same input:

| model | output RMS | peak |
|---|---|---|
| f16  | 0.128101 | 0.612 |
| q8_0 | 0.127733 | 0.595 |
| q4_k | 0.126898 | 0.583 |

That matters because q8_0 is essentially lossless (~0.9998 cosine), so silence there can't be a precision effect — it has to be a code path that only quantized weights take. Since it doesn't happen on CPU here, the most likely candidate is your execution backend.

Could you help with three things?

1. **`crispasr --version`** — the first line also lists the compiled backends, e.g. `[backends: cpu]`.
2. **Your OS and GPU** (macOS/Metal, Windows or Linux with CUDA, Vulkan, …), and whether you're using a release download or a local build.
3. **Re-run one failing command with `-ng`** (force CPU):
   ```
   crispasr -ng -m sidon-v0.1-q8_0.gguf -f in.wav --s2s --s2s-output upscaled_cpu.wav
   ```
[...]

PR #422, opened 2026-09-02
feat(mel-band-roformer): ggml-graph GPU port with fused single-graph forward (Change 176)

tilllt:
### Summary

GPU port of the mel-band-roformer separation backend (Refs #415): the forward pass is now built as ggml graphs instead of the host-side C++ loops, and a **fused single graph** runs band-split + the full 12-block time/freq transformer stack + the mask estimator on-device in one pass. Built cleanroom following the in-repo htdemucs graph pattern (PR #414), with chenmozhijin/BSRoformer.cpp (MIT) used strictly as a correctness oracle — no code copied; attribution noted in the file header.

### What changed

- **Phase 1/2:** band-split and the RoFormer transformer blocks as ggml graphs, byte-identical to the validated CPU path (same builder functions, no duplication).
- **Phase 5 (fused):** one graph per segment for the whole forward — removes the 13 host↔GPU roundtrips per segment (~740 MB PCIe traffic per layer) and moves the mask estimator off the CPU.
- **Segmenting:** Demucs-style 10 s segments with 25% overlap and triangular weighting for long audio (the attention matrix is O(T²·bands·heads); the unsegmented path OOMs beyond ~10 s).
- **Path selection** mirrors the htdemucs-gates architecture you established in #414 (`src/mel_band_gates.h`, pure `resolve()` + decision-table unit tests in `tests/test-mel-band-gates.cpp`, 9 cases): env unset = AUTO, `=0`/`=1` force either way, env beats caller intent in both directions, `FUSED=1` alone implies the graph path. Defaults stay on the validated CPU path; the AUTO upgrade to fused-GPU happens exactly when a real GPU backend is present and permitted.

### Verification

| Check | Result |
|---|---|
| CPU parity (golden10s, all stages) | cos=1.000000 (band_split, layer0_time, layer0_freq) |
| Fused graph vs CPU chain (7,923,916 elems) | cos=1.000000, max_abs 1.08e-4 (F16 noise level) |
| GPU parity, CUDA0 (same harness) | cos=1.000000 — fused_mask_raw PASS |
| Per-layer vs fused stems (30 s clip, vocals+other) | cos=0.99999999 (16-bit LSB only) |
| Path-selection unit tests | 9/9 passed (30 assertions) |
| `mel_band_roformer parity:` | PASS |

### Measured (RTX 3090 Ti, golden10s / 30 s clip / full track)

| Path | 30 s clip | RTF (full 359 s track) |
|---|---|---|
| CPU (legacy, validated reference) | ~28 min | ~57 |
| GPU per-layer graphs (13 roundtrips/segment) | 4:47 | ~10 |
| **GPU fused single graph** | **2.57 s** | **0.076** (~112× vs per-layer; 359 s song in 27 s) |

### Review questions

- **Default/AUTO semantics:** this PR ships the same AUTO gate pattern as #414 (fused-GPU exactly when a real GPU backend exists, legacy CPU path otherwise, no behavior change on CPU-only hosts). If you'd rather keep the mel-band graph path strictly opt-in for now (as the very first #414 iteration did), say so and I'll flip the AUTO default off.
- **Segment length:** default 10 s is hardcoded (`params.segment_seconds <= 0`); `CRISPASR_MELBAND_SEG_S` overrides. Happy to expose it as a CLI flag if you want it user-facing.

### Notes

- Refs #415 (planning/acceptance discussion) — do not auto-close on merge.
- Env docs added in `docs/environment-variables.md` (CRISPASR_MELBAND_GGML / GPU / FUSED / SEG_S / NO_SEGMENT), same style as the HTDemucs section.
- —
- *Drafted with assistance from Hermes, an AI agent (Nous Research Hermes Agent), on behalf of the contributor tilllt.*

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -864,10 +864,12 @@ All three optimisation gates are output-equivalent: the per-stage diff reports
   bisection arm. The full decision table is unit-locked in
   tests/test-mel-band-gates.cpp.
 - `CRISPASR_MELBAND_SEG_S` — override the Demucs-style segment length in
-  seconds (`params.segment_seconds`; <=0 → default 10 s). The attention
-  matrix is O(T²·bands·heads), so long inputs are split into ~10 s segments
-  with 25% overlap and a triangular weight (bounds VRAM; the unsegmented
-  whole-buffer path OOMs on any clip beyond ~10 s).
+  seconds (`params.segment_seconds`; <=0 → the checkpoint's trained
+  `chunk_size` from GGUF metadata, 8 s Kim fallback — do NOT expect 10 s: the
+  earlier hardcoded 10 s ran RoPE 25% past the trained window, review #422).
+  The attention matrix is O(T²·bands·heads), so long inputs are split into
+  segments with 25% overlap and a triangular weight (bounds VRAM; the
+  unsegmented whole-buffer path OOMs on any clip beyond ~10 s).
 - `CRISPASR_MELBAND_NO_SEGMENT` — process the whole track in one pass instead
   of the segmented overlap-add schedule (A/B against the old behaviour).
 
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -844,7 +844,32 @@ All three optimisation gates are output-equivalent: the per-stage diff reports
 
 ### Mel-Band RoFormer (source separation)
 
-- `CRISPASR_MBR_PROFILE`
+- `CRISPASR_MBR_PROFILE` — print a per-stage wall-time breakdown of one forward
+  pass (stft+pack / band_split / run_time / run_freq / mask_est / synthesize).
+- `CRISPASR_MELBAND_GGML` — run the ggml graph path instead of the legacy CPU
+  path. Since the Change-176 graph port the default is **AUTO**: ON exactly
+  when a real GPU backend is present and permitted (the fused single graph
+  measured ~112x faster than the per-layer graphs — RTF ~0.09 vs ~10 on an
+  RTX 3090 Ti), OFF on CPU-only hosts. `=1`/`=0` force either way.
+- `CRISPASR_MELBAND_GPU` — GPU permission (CUDA > Metal > Vulkan). Default
+  AUTO follows the caller's use_gpu (CLI default on); an explicit `=0`/`=1`
+  beats the caller in both directions — so `=0` genuinely opts out even
+  though the CLI defaults `use_gpu=true` (#414 review semantics). On GPU-less
+  hosts everything resolves to the CPU path regardless.
+- `CRISPASR_MELBAND_FUSED` — single fused graph: band-split + the full
+  time/freq transformer stack + mask estimator on-device in one graph, no
+  per-layer host↔device roundtrips (the measured-fastest path on GPU).
+  Default **AUTO**: ON with the GPU graph path, OFF otherwise. `=1` alone
+  implies the graph path it needs; `=0` on GPU keeps the per-layer-graph
+  bisection arm. The full decision table is unit-locked in
+  tests/test-mel-band-gates.cpp.
+- `CRISPASR_MELBAND_SEG_S` — override the Demucs-style segment length in
+  seconds (`params.segment_seconds`; <=0 → default 10 s). The attention
+  matrix is O(T²·bands·heads), so long inputs are split into ~10 s segments
+  with 25% overlap and a triangular weight (bounds VRAM; the unsegmented
+  whole-buffer path OOMs on any clip beyond ~10 s).
+- `CRISPASR_MELBAND_NO_SEGMENT` — process the whole track in one pass instead
+  of the segmented overlap-add schedule (A/B against the old behaviour).
 
 ### MeloTTS
 
--- a/examples/cli/crispasr_diff_main.cpp
+++ b/examples/cli/crispasr_diff_main.cpp
@@ -1307,6 +1307,11 @@ int main(int argc, char** argv) {
         // diff binary), so the backend shipped with no per-stage evidence.
         return mel_band_roformer_diff(model_path.c_str(), ref_path.c_str(), audio_path.c_str(), /*verbosity=*/2);
     }
+    if (backend_name == "mbr-parity" || backend_name == "mel-band-parity") {
+        // Change 176 Phase 1: standalone graph-vs-CPU band-split parity on a
+        // wav (no Python fixture needed). model_path = GGUF, audio_path = wav.
+        return mel_band_roformer_parity(model_path.c_str(), audio_path.c_str(), /*verbosity=*/2);
+    }
     if (backend_name == "rvc" || backend_name == "rvc-svc") {
         // model_path = rvc GGUF, ref_path = dump from tools/rvc_torch_parity.py.
         // Input-aligned AND noise-aligned: the reference carries input_phone,
--- a/examples/cli/crispasr_separate_cli.cpp
+++ b/examples/cli/crispasr_separate_cli.cpp
@@ -97,6 +97,8 @@ int crispasr_run_separate(const whisper_params& params) {
             // #414: forward the CLI's GPU intent — zero-initialized default
             // params carry use_gpu=false, which silently pinned separation to
             // CPU even on GPU builds (the moonshine "encoder gap" trap).
+            // Graph/fused/segment toggles come from default_params + the
+            // CRISPASR_MELBAND_* env overrides (Change 176), resolved in init.
             auto mp = mel_band_roformer_default_params();
             mp.use_gpu = params.use_gpu;
             mp.n_threads = params.n_threads;
--- a/models/convert-mel-band-roformer-to-gguf.py
+++ b/models/convert-mel-band-roformer-to-gguf.py
@@ -56,6 +56,12 @@ def _load_config(model_dir: Path):
     m.setdefault("stft_n_fft", a.get("n_fft", 2048))
     m.setdefault("stft_hop_length", a.get("hop_length", 441))
     m.setdefault("stft_win_length", a.get("n_fft", 2048))
+    # chunk_size = the inference window the checkpoint was TRAINED on (Kim
+    # vocals: 352800 = 8.0 s @ 44100). The runtime defaults segmentation to it;
+    # a 10 s default would extrapolate RoPE 25% past training (PR #422 review).
+    # Some configs keep it under `audio`; default 0 = omit the KV entirely so
+    # old runtimes fall back to their own 8 s default.
+    m.setdefault("chunk_size", a.get("chunk_size", 0))
     return m
 
 
@@ -153,6 +159,8 @@ def main():
     w.add_uint32("mel-band-roformer.stft_hop_length", hop)
     w.add_uint32("mel-band-roformer.stft_win_length", win)
     w.add_uint32("mel-band-roformer.stft_normalized", 1 if cfg.get("stft_normalized", False) else 0)
+    if cfg.get("chunk_size"):
+        w.add_uint32("mel-band-roformer.chunk_size", int(cfg["chunk_size"]))
     w.add_string("mel-band-roformer.instruments_json",
                  json.dumps(cfg.get("training", {}).get("instruments", ["vocals", "other"])
                             if isinstance(cfg.get("training"), dict) else ["vocals", "other"]))
--- a/src/mel_band_gates.h
+++ b/src/mel_band_gates.h
@@ -1,0 +1,62 @@
+// mel_band_gates.h — path selection for mel-band-roformer (Change 176),
+// pure and unit-testable. Mirrors the htdemucs_gates.h pattern from
+// PR #414 (src/htdemucs_gates.h): the same three coupled decisions,
+// resolved ONCE per init from the environment, the caller's use_gpu intent,
+// and whether a REAL (non-CPU) GPU backend exists.
+//
+//   use_graph   — run the ggml graph path instead of the legacy CPU path
+//   use_fused   — single fused graph (band_split + transformer stack + mask
+//                 estimator on-device, "no roundtrips"); only meaningful when
+//                 use_graph is set
+//   gpu_backend — place the graph on the GPU backend
+//
+// Measured facts the AUTO defaults encode (RTX 3090 Ti, 2026-09-02): the CPU
+// path is RTF ~57 (10 s clip = 9m35s); per-layer graphs on GPU run the same
+// clip in 4:47; the FUSED single graph in 2.6 s (~112x faster than per-layer,
+// RTF ~0.09 on a 359 s song incl. iSTFT). Per-layer graphs still pay a
+// host<->device roundtrip per layer and keep the CPU mask estimator, so AUTO
+// picks graph+fused+GPU exactly when a real GPU is present and permitted,
+// and the plain CPU path otherwise. Never a slower-than-before configuration.
+//
+// Env semantics (each var: unset = AUTO, "0" = force off, else force on):
+//   CRISPASR_MELBAND_GPU    — GPU permission; explicit value beats the
+//                              caller's use_gpu in BOTH directions (expert
+//                              override; the CLI forwards params.use_gpu, so
+//                              "params wins" would make GPU=0 dead — the
+//                              #414 review catch, fix 208b2d59)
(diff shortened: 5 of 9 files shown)


CrispStrobe:
Reviewed in depth. This is strong work — the cleanroom discipline, the reuse of `mbr_band_split_build` so the standalone and fused paths cannot structurally diverge, the residual-on-`x`-not-`xn` detail in the attention block, and the decision to mirror the #414 gates pattern rather than invent a new one are all exactly right. I traced the attention layout by hand (`q4/k4/v4` reshape → permute → `mul_mat` → per-head gate broadcast → `reshape_3d` back to `inner`) and the axis algebra is correct at every step, which is consistent with your cos=1.0 vs the CPU reference.

Below: which CI reds are actually yours (only one of three), then compliance, then the substantive findings. Two of them I think are worth acting on before merge; the rest are notes.

---

## 1. CI: two of the three reds are MINE, not yours

Please don't chase these:

- **`linux-backend-dl`** — ``undefined reference to `ggml_backend_is_cpu` `` comes from `src/mt3.cpp` at your merge base (`90b354aa`), not from your diff. Your code correctly uses `core_cpu_backend::is_cpu` throughout. I fixed it on main in `e77d0434`.
- **`linux-x86_64`** — `RESULT: FAIL (Go cgo LDFLAGS drift)`, also mine, from the MT3 merge. Fixed on main in `4c4d5f3b`.

**A rebase on current main clears both.** Sorry for the noise — you inherited my breakage.

- **`clang-format` is genuinely yours**: ~20+ violations in `src/mel_band_roformer.cpp`. `tools/format.sh --fix src/mel_band_roformer.cpp` handles it. Most of it is the new `} else {` block in `run_forward` whose body kept its old indentation — worth an eyeball after the auto-fix, since that block is now nested one level deeper and currently reads as if it were still top-level.

## 2. `docs/contributing.md` compliance

Walked the checklist. **It complies**, and a couple of the traps don't apply here:

- §367 *append-only ABI structs — update EVERY hand-written mirror*: does **not** apply. `mel_band_roformer_params` is internal (`src/` only, not in the C ABI surface, not mirrored in any binding), and the C ABI constructs it through `mel_band_roformer_default_params()`. Adding `segment_seconds` / `no_segment` is safe. I checked rather than assumed.
- §Docs: env vars documented in `docs/environment-variables.md` in the established style. Good.
- §Build targets / unit test registration: `tests/CMakeLists.txt` wiring matches the `test-htdemucs-gates` precedent, correct labels and `TEST_SPEC`.
- The one doc that is now stale is `docs/mel-band-roformer/PLAN.md` — see finding 4, which is the substantive reason to touch it.

## 3. Unit tests: the gates test is good — I verified it independently

I didn't take the 9 cases on trust. I extracted `mel_band_gates.h`, wrote a *separate* oracle from the documented semantics alone (unset = AUTO, `"0"` = off, else on), and swept **all 108 combinations** of `{unset,"0","1"}³ × caller_use_gpu × have_real_gpu`:

```
checked 108 combinations, 0 mismatches
```

All three invariants hold everywhere: `use_fused ⇒ use_graph`, `gpu_backend ⇒ use_graph`, `gpu_backend ⇒ have_real_gpu`. The decision table is correct.

Two notes, neither blocking:

- `atoi()` means a non-numeric truthy value silently means *off*: `CRISPASR_MELBAND_GPU=true` → `gpu_backend=0`, which contradicts the header's documented "else = force on". This is **inherited verbatim from `htdemucs_gates.h`**, so it's a repo-wide wart and not yours to fix here — I'm noting it so it's on the record.
- `mel_band_gates.h::resolve()` is a byte-for-byte copy of `htdemucs_gates.h::resolve()` (only a trailing comment differs). Two copies is tolerable; the third will not be. Worth hoisting to a shared `src/core/backend_path_gates.h` at some point — a follow-up, not a merge blocker.

## 4. ⚠️ The segment length disagrees with the model's own trained `chunk_size`

This is my main technical finding.

`docs/mel-band-roformer/PLAN.md` records the Kim vocals config as:

```
sample_rate 44100   chunk_size 352800   ...   stft_hop_length 441
```

`chunk_size 352800` = **8.0 s** — that is the window the checkpoint was trained and is intended to be inferred at. The PR hardcodes **10 s** (`seg_s > 0 ? seg_s : 10`), which at hop 441 is 1000 STFT frames against the trained 800. The time-transformer's RoPE positions therefore run 25% past anything seen in training, on every segment, for all long audio. That's extrapolation, and it's the kind of thing that shows up as subtle artifacts rather than an obvious failure — which is why none of the current evidence would catch it.

`chunk_size` is currently **not** carried into the GGUF at all (`grep chunk_size` finds nothing in the converter, `src/`, or the reference dumper). Suggested fix, in preference order:

1. Plumb `chunk_size` through the converter into GGUF metadata and default `segment_seconds` from it — correct for any future checkpoint, not just Kim.
2. Failing that, change the hardcoded fallback from 10 to 8 s and say in a comment that it's the Kim `chunk_size`, so the next checkpoint's mismatch is at least greppable.

`CRISPASR_MELBAND_SEG_S` already gives the escape hatch, so this is about the *default* being right.

## 5. ⚠️ The 30 s arm cannot detect an overlap-add error

In the verification table:

| Per-layer vs fused stems (30 s clip) | cos=0.99999999 |

Segmentation kicks in above 10 s, and it lives in `mel_band_roformer_separate`, *above* the gates — so on a 30 s clip **both** arms are segmented identically. Any error in the windowing or overlap-add cancels between them exactly. That arm is real evidence about per-layer-vs-fused; it is structurally incapable of being evidence about segmentation.

And the other arm doesn't cover it either: `golden10s` is exactly 441000 samples, and the guard is `n_samples <= seg_len`, so the fixture takes the whole-buffer path and **never segments**. Net: the single largest behavioural change in the PR has no correctness evidence behind it.

**The good news — I tested the machinery myself and it's correct.** I replicated the schedule (triangular weight, `stride = 0.75·seg_len`, accumulate, normalise by `sum_weight`) with the model as identity, and swept ~200 combinations of track length and segment length, including every boundary case around `n_samples = seg_len + 1`:

```
TOTAL HOLES ACROSS SWEEP: 0        max_recon_err = 2.4e-07 (float noise)
```

No coverage holes, no `sum_weight == 0` division, exact reconstruction including the final partial segment. So I'm **not** claiming a bug — the arithmetic is sound. What's missing is evidence that segmented output matches the *reference model's* behaviour on long audio, which is a different question from whether the adder works, and which finding 4 makes more pressing.

## 6. Live tests — what I'd want before merge

Unit tests are fine as they stand. The live gap is the one worth closing:

- **End-to-end against the Python reference on a >10 s clip.** Everything today anchors the graph path to this file's CPU path, which is a valid transitive chain for `band_split`/`layer0`/`fused_mask`. But no arm anchors *segmented* output to the reference. Running upstream `bs-roformer==0.3.10` on the same 30 s clip and comparing stems would close it — and would also settle finding 4 empirically, since the reference chunks at its own `chunk_size`.
- **A cheap regression guard for the schedule**, in the spirit of the gates test: assert no coverage holes and exact identity reconstruction over a few `(n_samples, seg_len)` pairs including `seg_len + 1`. It passes today; it's there so it keeps passing. Happy to contribute the file if you'd rather not.

## 7. Notes

- **The fused path doesn't reach library/server consumers.** `src/crispasr_c_api.cpp:2922` inits with a bare `mel_band_roformer_default_params()` and never sets `use_gpu`, so for bindings and the server the gates resolve to the legacy CPU path — the same "encoder gap" shape as the #414 CLI catch you cite. `CRISPASR_MELBAND_GPU=1` does rescue it, so it's not a defect, but the headline speedup currently lands only for `crispasr --separate`. Worth deciding whether to forward intent there too.
- **`mel_band_roformer_parity` inits from `default_params()` (`use_gpu=false`).** So a bare `crispasr-diff mbr-parity` compares *CPU*-graph against CPU-reference even on a GPU box; the GPU parity result requires `CRISPASR_MELBAND_GPU=1`. The init line prints the backend name, so it's visible — but it's an easy way for a future runner to record a "GPU PASS" that was a CPU run. A one-line note in the harness output or the header comment would inoculate it.

## Answers to your two review questions

- **AUTO semantics:** keep AUTO as you have it. It matches #414, CPU-only hosts are unaffected, and the fallbacks are conservative. No objection.
- **Segment length as a CLI flag:** yes, but I'd prioritise finding 4 (default from `chunk_size`) over exposing the flag. The flag is nice-to-have; the default being 25% past the trained window is the part that affects everyone who never sets it.

---

**Verdict: changes requested, but small.** Blocking: `clang-format`, and a rebase to clear my two reds. Strongly recommended before merge: the `chunk_size` default (finding 4). The rest are notes and follow-ups. The core port is careful and I'd be glad to see it land.

tilllt:
Thanks for the deep review — all three sections addressed. The branch has been rebased onto current main (now `0` commits behind, head `fb91d3991`) and the clang-format pass is in; both of your CI reds are cleared along with mine (28/28 green on the rebased head).

**Finding 4 (chunk_size) — fixed, option 1 as you preferred:**

- `models/convert-mel-band-roformer-to-gguf.py` now carries the config's `chunk_size` into GGUF metadata as `mel-band-roformer.chunk_size` (read from the `model` or `audio` section; omitted entirely when absent, so older checkpoints keep working)
- the loader reads it into hparams and the segment default is now `chunk_size / sample_rate`, i.e. the trained 8.0 s window for the Kim vocals checkpoint — no more RoPE extrapolation on the default path
- fallback is 8 s (not 10 s) for GGUFs that predate the metadata, with a comment naming the Kim chunk so the next mismatch is greppable
- `CRISPASR_MELBAND_SEG_S` / `params.segment_seconds` still override; env doc + header comment updated

**Finding 7 (notes):**

- `mel_band_roformer_parity` header now states that bare runs init from `default_params()` (CPU) and that a GPU run requires `CRISPASR_MELBAND_GPU=1`, with the printed backend name as the ground truth — so a future runner can't record a GPU PASS from a CPU run.
- The c_api/`crispasr_c_api.cpp:2922` use_gpu gap: agreed it's the #414 encoder-gap shape. Rather than guess at the intent-forwarding semantics for library consumers, I'd like your call: forward `use_gpu` through the C ABI init (additive param/default), or keep the env override as the documented path for now? Happy to implement whichever you prefer in a follow-up.

**Finding 5 (segmentation evidence):** working on it. I'll run the >10 s clip through the upstream reference and compare stems; if that stalls on environment setup I'll take you up on the regression-guard offer.

**Your two notes (atoi truthiness, gates-header copy):** acknowledged, both tracked as follow-ups on my side — agreed neither belongs in this PR.

tilllt:
Finding 5 — segmentation evidence is now in. I ran the 30 s clip (`kibox_long_30s.wav`, exactly 1,323,000 samples) through two independent implementations:

- **Reference:** upstream lucidrains bs-roformer (`bs-roformer==0.3.10`) with the Kim weights (`MelBandRoformer.ckpt`, loaded with `missing=0, unexpected=0` against the documented Kim config), chunked at the trained `chunk_size` 352800 with the same Demucs OLA schedule CrispASR uses (25% overlap, triangular weight, accumulate + normalize).
- **CrispASR:** the PR code, CPU path, `CRISPASR_MELBAND_SEG_S=8` (equivalent to the new chunk_size default), 5 segments.

Result on the vocals stem:

```
cos channel 0: 0.99990648
cos channel 1: 0.99991119
cos global   : 0.99990892
RMS diff     : 0.000796   (signal RMS 0.0588)
```

So the segmented output is anchored to the real model on a >10 s clip — the overlap-add path no longer has zero evidence behind it. Honest scope note: both sides used the same 8 s / 25% / triangular schedule, so this validates numerical parity of the segmented implementation against the reference across segment boundaries; it does not independently validate the schedule *parameters* against Kim's own inference defaults. That choice is now at least model-faithful (default = the checkpoint's trained chunk_size per finding 4), and your own ~200-combination sweep already covered the adder arithmetic.

Also: the branch was rebased again onto the latest main (c59c6154b; sidon #416 landed after my earlier rebase — the only conflict was `tests/CMakeLists.txt` where both test blocks now coexist). Current head `47ae57a38` includes the chunk_size fix; CI is green except the last job still running.

The regression-guard offer from your review is still welcome if you'd like to contribute it — the E2E above was the one-off evidence, a cheap weight-free guard would keep it honest going forward.

CrispStrobe:
**Merged.** This landed on `main` as `c80cfeed`, with all seven of your commits preserved under your authorship — verified each one is an ancestor of `main`. Closing manually because the merge was done locally (this repo's convention) and GitHub doesn't auto-detect that.

Thank you for this. The care shows in places that are easy to skip: reusing `mbr_band_split_build` so the standalone and fused paths *cannot* structurally diverge, getting the residual onto the un-normalised `x` rather than `xn`, and following the #414 gates pattern instead of inventing a new one. I traced the attention layout by hand — `q4/k4/v4` reshape → permute → `mul_mat` → per-head gate broadcast → `reshape_3d` back to `inner` — and the axis algebra is correct at every step, which is consistent with your cos=1.0.

**On the review round:** you fixed clang-format (re-verified here: 0 violations), and for the segment length you took the thorough option rather than the cheap one — converter → GGUF metadata → loader → default, with the 8 s fallback for older files. I traced that plumbing rather than trusting the commit message: `_load_config()` returns the dict that merges `chunk_size` from either the `model` or `audio` section, and the writer reads that same dict, so the KV genuinely gets written. That was the right call and it will keep working for checkpoints that aren't Kim.

**Two things I changed on top, in `a8c0327e`** — both small, neither a criticism of the approach:

1. The trained chunk is now applied **in samples** rather than round-tripped through integer seconds. `chunk_size / sample_rate * sample_rate` is exact for Kim (352800 / 44100 = 8) but truncates any checkpoint whose chunk isn't a whole number of seconds — 344400 (7.81 s) would floor to 7 and drop 0.81 s of trained context from every segment. Same class of defect your fix existed to remove, just smaller.
2. The default had grown four sources (env, params, GGUF chunk, fallback) resolved inline, so I moved it to a pure `mel_band_gates::resolve_segment_len()` next to your path gates and unit-locked it — the suite is now 46 assertions / 14 cases. The odd-chunk case asserts `344400 != (344400/44100)*44100`, so it fails if anyone reintroduces the round-trip. It also locks two behaviours that were implicit: a junk env value is ignored rather than obeyed (`atoi("abc") == 0`, and obeying that would let a typo silently disable the trained default), and an absurd override returns 0 instead of overflowing into a negative the caller would read as "no segmentation".

Also cleared the `-Wunused` warnings the port introduced, and documented that `mel_band_roformer_diff()` deliberately ignores its `audio_wav` argument — it reads the reference GGUF's own `input_audio`, which is the *right* design (both sides then see byte-identical samples, so a decode difference can't be mistaken for a model difference), but silently disregarding a path argument is worth stating.

**One review finding is still open**, recorded in the merge commit so it isn't mistaken for settled: segmentation has no reference-anchored evidence. Your 30 s arm compares per-layer vs fused, but segmentation sits *above* the gates, so both arms are segmented identically and an overlap-add error would cancel between them; and `golden10s` is exactly at the threshold (`n_samples <= seg_len`) so it never segments. I did verify the overlap-add **machinery** independently — replicated the schedule with the model as identity across ~200 length/segment combinations including every `seg_len + 1` boundary: no coverage holes, no `sum_weight == 0` division, reconstruction exact to 2.4e-07. **So the adder arithmetic is sound and this is not a bug report.** What's missing is an end-to-end comparison against the Python reference on a >10 s clip, which needs a GPU; that's queued on our side, not yours.

Two notes for whenever you're next in this code, neither blocking:

- The fused path doesn't reach library or server consumers — `src/crispasr_c_api.cpp` inits with bare default params and never sets `use_gpu`, so the gates resolve to the legacy CPU path there. `CRISPASR_MELBAND_GPU=1` rescues it. Same "encoder gap" shape as the #414 CLI catch you cite, so you may want to forward intent there too.
- `mel_band_gates::resolve()` is currently a byte-for-byte copy of `htdemucs_gates::resolve()`. Two copies is fine; the third won't be. Worth hoisting to a shared header eventually.

On your two questions: AUTO semantics stay as you have them — they match #414, CPU-only hosts are unaffected, and the fallbacks are conservative. And the segment length as a CLI flag is welcome whenever you feel like it; the default being right mattered more, and that's done.

The measured numbers deserve restating: 28 min → 2.57 s on a 30 s clip, RTF ~0.076 on a full track. Thanks again — and sorry for the two red checks at the start that turned out to be my breakage on `main` rather than anything in your branch.

